PR #356, opened 2020-05-21
Kalman Filter

aewebb80:
The current version of the filter will track two instances frame by frame. The function requires the frames and the instance count. 

The function may be imported by:
`from sleap.nn.kalman_filter import filter_frames`

Basic usage (within TrackCleaner) :
`filter_frames(frames=frames, instance_count=self.instance_count)`

The function also includes an optional argument - `keep_non_tracked`. This will keep instances that were unable to be tracked (on new Tracks). This should likely be configured by argparse, but felt it was best to leave that to you.


buildbot-princeton:

Issue #390, opened 2020-08-19
Validation fraction config not saved


iteal:
Hi! I am trying to use SLEAP without the GUI and I looked at this piece of code in sleap.nn.training (DataReaders.from_config)

```python
elif isinstance(validation, float):
    validation_labels = labels_config.validation_fraction # This line has no effect
```
The labels config is not updated after passing a validation fraction has an input.

PR #391, opened 2020-08-19
Update correctly the configuration for validation fraction

iteal:
The validation fraction set by a user was not saved in the config. Modified a line with no effect.


buildbot-princeton:

iteal:
for issue #390

talmo:
Thanks for spotting this! The validation fraction is still used appropriately to split the data, but you're right that it wasn't getting updated in the config.

PR #399, opened 2020-09-08
Fix type warning in run_tracker

iteal:
run_tracker can accept a BaseTracker which is more generic than a Tracker.
I get a warning when doing this:

tracker = Tracker.make_tracker_by_name() # this returns a BaseTracker
frames = run_tracker(frames=frames, tracker=tracker) # Warning, run_tracker expects a "Tracker"

### DIFF
--- a/sleap/nn/tracking.py
+++ b/sleap/nn/tracking.py
@@ -1059,7 +1059,7 @@ class TrackCleaner:
         connect_single_track_breaks(frames, self.instance_count)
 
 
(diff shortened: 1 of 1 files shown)


buildbot-princeton:

iteal:
By the way, which "black" version are you using? The version is not pinned in dev_requirements.txt, and I know different versions give different results, so I can't be sure if the formatting is correct.

talmo:
LGTM.

Also I'm on [black 19.10b0](https://black.readthedocs.io/en/stable/change_log.html#id1), but it doesn't look like there's breaking changes in 20.x.

PR #424, opened 2020-12-03
SLEAP 1.1.0 and TF2.3 support

talmo:


sleap/instance.py
```diff
@@ -1141,6 +1141,19 @@ def __delitem__(self, index):
...
+    def __repr__(self) -> str:
+        """Return a readable representation of the LabeledFrame."""
+        return (
+            f"LabeledFrame(video={type(self.video.backend).__name__}"
+            f"('{self.video.filename}'), "
+            f"frame_idx={self.frame_idx}, "
+            f"instances={len(self.instances)})"
+        )
+
+    def __str__(self) -> str:
```
ariematsliah, line 1153: This might be redundant (I think str defaults to repr when not implemented)

talmo: Ack -- this was because I was overriding the default attrs behavior which creates a method for each. Just turning it off with the attrs decorator now.

PR #425, opened 2020-12-08
Fix CPU thrashing

ariematsliah:
On my 5 year old MacBook PRO CPU is constantly at >100% when sleap-label is running (even in idle mode).
- Ran profiler and found that multiple threads were running checks in busy loops (using timer)
- Set all timers to run with 20 milisec intervals (50 Hz)
Now CPU usage down to 0% when idle

### DIFF
--- a/sleap/nn/monitor.py
+++ b/sleap/nn/monitor.py
@@ -236,10 +236,10 @@ class LossViewer(QtWidgets.QMainWindow):
             self.zmq_ctrl = self.ctx.socket(zmq.PUB)
             self.zmq_ctrl.bind("tcp://127.0.0.1:9000")
 
-        # Set timer to poll for messages
+        # Set timer to poll for messages every 20 milliseconds
         self.timer = QtCore.QTimer()
         self.timer.timeout.connect(self.check_messages)
-        self.timer.start(0)
+        self.timer.start(20)
 
     def stop(self):
         """Action to stop training."""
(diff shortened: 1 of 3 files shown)


buildbot-princeton:

sleap/nn/monitor.py
```diff
@@ -423,7 +423,7 @@ def test_point(x=[0]):
```
talmo, line 423: This whole section is just a convenience entrypoint for development.

Do you mind moving this into a test? (with the 20ms timeout)

talmo (changes requested):
LGTM and happy to see the performance improvement in the training monitor!

Just convert that `__main__` section into a test to make sure the monitor can get launched and updated with a point and the timer at >0ms.

ariematsliah:
> LGTM and happy to see the performance improvement in the training monitor!
> 
> Just convert that `__main__` section into a test to make sure the monitor can get launched and updated with a point and the timer at >0ms.

Just to clarify the goal is improving the performance of the main GUI app (sleap-label). Then I was just looking for other places where the timers were configured without a sufficient interval and found the training monitor.

sleap/nn/monitor.py
ariematsliah: Ro you mean moving this logic to a unit test?

ariematsliah:
Merging based on discussion offline

PR #433, opened 2020-12-13
Small fix: correctly pass arguments

ariematsliah:
One of the calls to read_frames was assigning a 'video' to 'skeleton'.

Thoughts on adding mypy in CI?

### DIFF
--- a/sleap/io/format/deeplabcut.py
+++ b/sleap/io/format/deeplabcut.py
@@ -69,7 +69,7 @@ class LabelsDeepLabCutCsvAdaptor(Adaptor):
     def read(
         cls, file: FileHandle, full_video: Optional[Video] = None, *args, **kwargs,
     ) -> Labels:
-        return Labels(labeled_frames=cls.read_frames(file, full_video, *args, **kwargs))
+        return Labels(labeled_frames=cls.read_frames(file=file, full_video=full_video))
 
     @classmethod
     def make_video_for_image_list(cls, image_dir, filenames) -> Video:
@@ -93,8 +93,6 @@ class LabelsDeepLabCutCsvAdaptor(Adaptor):
         file: FileHandle,
         skeleton: Optional[Skeleton] = None,
         full_video: Optional[Video] = None,
-        *args,
-        **kwargs,
     ) -> List[LabeledFrame]:
         filename = file.filename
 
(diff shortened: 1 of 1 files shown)


sleap/io/format/deeplabcut.py
```diff
@@ -69,7 +69,7 @@ def does_write(self) -> bool:
     def read(
         cls, file: FileHandle, full_video: Optional[Video] = None, *args, **kwargs,
     ) -> Labels:
-        return Labels(labeled_frames=cls.read_frames(file, full_video, *args, **kwargs))
```
talmo, line 72: Are we sure that skeleton doesn't ever get passed in here through the `**kwargs`?

talmo (approved):
Nice catch! LGTM, but see comment below.

mypy: Yes, please feel free to set this up in the CI workflow (`.github/workflows/ci.yml`) in another PR :). I imagine we'll have to do a lot of fixes to get that to pass.

sleap/io/format/deeplabcut.py
ariematsliah: not sure if I understand the question - the 'read_frames' method does not access args / kwargs

ariematsliah:
re mypy sounds good - for now dropped a 'todo' under our project

sleap/io/format/deeplabcut.py
ariematsliah: but the method read_frames is not reading anything from kwargs, so this is a no-op right?

talmo: It's not that `read_frames` is using the `**kwargs`, it's that it's using `skeleton`. Previously, if `kwargs = {"skeleton": [...]}` in `read()`, that would pass through to the named keyword arg in `read_frames()`. With this change, you cannot pass a skeleton to `read_frames()` through `read()`.

I don't think this is a serious use case, but to ensure it is equivalent to the previous behavior, `read()` should be:

```py
@classmethod
def read(cls, file: FileHandle, full_video: Optional[Video] = None, skeleton: Optional[Skeleton] = None, *args, **kwargs) -> Labels:
    return Labels(labeled_frames=cls.read_frames(file=file, full_video=full_video, skeleton=skeleton))
```

See what I'm saying?

ariematsliah: Thanks for the explanation - adding them back

talmo (approved):
LGTM

PR #434, opened 2020-12-13
Update gitignore for Mac OS

ariematsliah:


talmo:
LGTM.

Also realizing that the required checks (CI tests) won't run for PRs without any commits to `sleap/*` or `tests/*`... Let's just force merge these for now until we can think of a smarter branch protection rule...

Issue #209, opened 2019-10-07
Datagen fails on videos with different dimensions


ntabris:
We gather annotated frame images from all videos and then try to combine them into single ndarray here:

https://github.com/murthylab/sleap/blob/develop/sleap/nn/datagen.py#L137

If videos have different dimensions, this results in `ValueError('all input arrays must have the same shape')`.

Sample file at:
`/tigress/MMURTHY/Sama/X079_sleap/abdus-saboor/grant/nat.json`

talmo:
Fixed in #440 (finally)

PR #440, opened 2020-12-22
Multi-size videos in data pipelines

talmo:
Support for datasets with videos with different image shapes in the data pipeline

Addresses the longstanding #209.

### DIFF
--- a/sleap/nn/config/data.py
+++ b/sleap/nn/config/data.py
@@ -76,6 +76,13 @@ class PreprocessingConfig:
             max stride (typically 32). This padding will be ignored when instance
             cropping inputs since the crop size should already be divisible by the
             model's max stride.
+        resize_and_pad_to_target: If True, will resize and pad all images in the dataset to match target dimensions.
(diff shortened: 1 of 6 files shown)


talmo:
The approach here should be:

1. Determine target height/width (max of all videos or user-specified)
2. Resize longest side to target height/width
3. Pad (bottom/right) the short side to target width/height

Other notes:
- This metadata will need to be saved in the config so these steps can be reproduced during inference time. This may not be strictly necessary since we can retrace for different sizes as long as each batch is from the same video, but we should test.
- This padding should be aware of/compatible with the padding done to account for model stride ([`sleap.nn.data.resizing.pad_to_stride`](https://github.com/murthylab/sleap/blob/master/sleap/nn/data/resizing.py#L33)/[`Resizer`](https://github.com/murthylab/sleap/blob/master/sleap/nn/data/resizing.py#L100)). Ideally they'd be done in the same step, but this gets complicated for multi-model approaches (topdown) that may require different padding.

PR #445, opened 2021-01-05
Inference fixes and cleanup

talmo:
This PR fixes a bug and cleans up the inference module.

- Top-down models that need to get converted to RGB would fail when using grayscale images since the number of channels of the centroid crops was being determined using their dynamic rather than static shape.
- Completed most of the docstrings and typing in the module and added a longer description on the APIs it exposes.

**Breaking:**
- `TopDownModel` is renamed to `TopDownInferenceModel` for consistency with `BottomUpInferenceModel` and `SingleInstanceInferenceModel`.
- `Predictor.predict()` no longer has `make_instances` kwarg. Instead, just leave `make_labels=True` and use `predictor.predict(x, make_labels=True).labeled_frames` to pull out a list of labeled frames since there's little overhead in instantiating a full `sleap.Label`s object.
- Prefixed several methods in the predictors to denote their private usage. The main interface to `Predictor` subclasses should be `predict()`.


sleap/nn/inference.py
```diff
@@ -1695,49 +1895,42 @@ def make_labeled_frames_from_generator(self, generator, data_provider):
...
+                dictionaries containing the raw arrays returned by the inference model.
 
-        return list(generator)
+        Returns:
+            A `sleap.Labels` with `sleap.PredictedInstance`s if `make_labels` is `True`,
+            otherwise a list of dictionaries containing batches of numpy arrays with the
+            raw results.
+        """
+        # Create provider if necessary.
+        if isinstance(data, sleap.Labels):
```
ariematsliah, line 1918: I know this is not new, but just curious how is this casting helpful here?

ariematsliah (approved):
LGTM
Great docs!

sleap/nn/inference.py
talmo: This is a convenience shortcut. Say you have some `sleap.Labels` you want to predict on. This saves an unnecessary step:
```py
predictor.predict(labels)
```
as compared to:
```py
predictor.predict(sleap.pipelines.LabelsReader(labels))
```

PR #451, opened 2021-01-08
[offset_regression] Add optional refinement heads to config and training for all models

talmo:
Offset refinement heads can now be specified through the config system and all model types can be trained with offset refinement heads


sleap/nn/config/model.py
```diff
@@ -28,11 +28,19 @@ class SingleInstanceConfmapsHeadConfig:
             results in confidence maps that are 0.5x the size of the input. Increasing
             this value can considerably speed up model performance and decrease memory
             requirements, at the cost of decreased spatial resolution.
+        offset_refinement: If `True`, model will also output an offset refinement map
```
ariematsliah, line 31: looks like a lot of the docstrings are shared, have you considered reusing them?
Something like
```
class Base(object):
    """
    Base Class docs here
    """
    pass

class ImplA(Base):
    """
    ImplA specific docs here
    """ 
    __doc__ += Base.__doc__
    pass
```
Or use some decorators for docs inheritance

ariematsliah (approved):

sleap/nn/config/model.py
talmo: Feel free to add a card to the to-do board -- there's a lot of places where we'll need to clean up :)

PR #452, opened 2021-01-08
[offset_regression] Add offset refinement to inference

talmo:
- Add routines for peak finding with offset refinement maps
- Inference classes can detect refinement maps from model outputs and use appropriate peak finding routines
- Add tests for model training with offset refinement heads
- Add trained models with offset refinement heads for inference tests


ariematsliah (approved):
LGTM

PR #484, opened 2021-02-22
Onboarding/workflow UX

talmo:
### Description
This PR adds a new set of GUI elements intended to guide users through the core workflow. This is intended to supplant wordy documentation guides on the website, keeping all GUI-only information actually within the GUI.

- [x] SLEAP icon
- [ ] Splash widget: New, Open, Inference
- [x] Main panel to the dock with tabs associated with each of the main labeling steps
- [x] Guide panel with auto-detection of each workflow step
- [ ] Simpler training widget to add as a tab
- [ ] Standalone inference widget
- [ ] Instance list panel (HUD is tricky to make editable/sortable)
- [ ] Tracks list (especially so you can edit track names)
- [ ] Proofreading mode (detect by seeing if project is all predictions with no suggestions and single video) -> shouldn't show core labeling workflow since this is the step where user fixes tracking
- [ ] Display tip panel with GIF demonstrating how to drag nodes, mark them as not visible, rotate, etc

### Types of changes

- [ ] Bugfix
- [x] New feature
- [ ] Refactor / Code style update (no logical changes)
- [ ] Build / CI changes
- [ ] Documentation Update
- [ ] Other (explain)

### Does this address any currently open issues?
No (?)

#### Thank you for contributing to SLEAP!
:heart:

PR #504, opened 2021-03-05
Revert "Pulling finished work from #484 into a separate PR"

ariematsliah-princeton:
Reverts murthylab/sleap#493


talmo:
For future reference:

I kept lots of the changes in, but reverted:
- Suggestion generation text (too bulky in the old panels-based layout)
- Suggestions UI elements
- All the `sleap.learning.configs` changes since it was failing to detect some of the default training profiles

PR #535, opened 2021-03-30
fix code samples in documentation

smidm:
The code samples to render properly have to be valid rst literal blocks (https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#literal-blocks). The most affected are multi-line code samples that are joined to a singe line, see https://sleap.ai/api/sleap.io.dataset.html.

The leading "   > " could be possibly removed to make copying the snippets easy.


smidm:
outcome: 

Screenshot from 2021-03-30 07-31-42

talmo:
LGTM! Thanks @smidm!

Issue #518, opened 2021-03-22
Colored, predicted instances shift when selected


olinesn:
Hi, I'm writing to report a GUI bug. I'm using SLEAP v1.1.1, but the bug has been present in previous versions.

Step 1: **Predict > Run Inference > Predict On: Current Frame > Run** gives 3 predicted instances:
Instance 1: Blue
Instance 2: Orange
Instance 3: Yellow
![image](https://user-images.githubusercontent.com/69451171/111985395-f30eaf80-8ae2-11eb-85d8-e979a267f9f2.png)

Step 2: Double click on a point from the blue instance, Instance 1. All instance numbers/colors shift, and this instance now become Instance 3 (Yellow).

![image](https://user-images.githubusercontent.com/69451171/111985425-f99d2700-8ae2-11eb-96bb-555ad3589f4a.png)

Is there a way to keep the instance numbers constant when selecting the predicted instances?
Thanks!

talmo:
Hi @olinesn,

Fixes this in v1.1.4 -- you can update with the instructions here: https://github.com/murthylab/sleap/releases/tag/v1.1.4

Cheers,

Talmo

PR #694, opened 2022-04-01
Hide predicted instances with hotkey

roomrys:
### Description
Hide predicted instances with same hotkey that hides user-labeled instances. Fix status bar message for hidden instances.

### Types of changes

- [x] Bugfix
- [x] New feature
- [ ] Build / CI changes
- [ ] Documentation Update
- [ ] Other (explain)

No.

- [x] Review the [guidelines for contributing](https://github.com/murthylab/sleap/wiki/Developer-Guide) to this repository
- [ ] Read and sign the [CLA](https://github.com/murthylab/sleap/blob/develop/sleap-cla.pdf) and add yourself to the [authors list](https://github.com/murthylab/sleap/blob/develop/AUTHORS)

:heart:


sleap/gui/app.py
```diff
@@ -1364,15 +1364,15 @@ def updateStatusMessage(self, message: Optional[str] = None):
                     )
                     message += " in video"
 
-            lf = self.state["labeled_frame"]
-            n_instances = 0 if lf is None else len(lf)
+            visible_instances = self.state["labeled_frame"].instances_to_show
```
talmo, line 1367: Is `self.state["labeled_frame"]` always defined? I thought it was initialized to `None`, which is why we were doing that check above.

If it's `None`, then doing `self.state["labeled_frame"].instances_to_show` will give you an attribute error saying that `NoneType` objects don't have an `instances_to_show` attribute.

sleap/gui/app.py
```diff
@@ -1364,15 +1364,15 @@ def updateStatusMessage(self, message: Optional[str] = None):
...
+            n_instances = 0 if visible_instances is None else len(visible_instances)
```
talmo, line 1368: I don't think `LabeledFrame.visible_instances` ever returns `None`? Again, if `self.state["labeled_frame"] == None`, you won't get a `None` back from an attribute that it doesn't have.

sleap/gui/app.py
```diff
@@ -1364,15 +1364,15 @@ def updateStatusMessage(self, message: Optional[str] = None):
...
             message += f"{spacer}Current frame: {n_instances} instances"
-            if not self.state["show instances"]:
+            if n_instances and not self.state["show instances"]:
```
talmo, line 1370: **Nit:** When using integers or other numeric types in a conditional expression, I find it more readable to explicitly do the binary comparison (`n_instances > 0`) than to rely on the implicit boolean casting (`bool(n_instances) == False` ~= `bool(0) == False`).

(PS: "nit" == a nit-picky suggestion based on style preference that is not mandatory.)

sleap/gui/app.py
talmo: One more thing: this doesn't require a change necessarily, but I'd discourage using the `LabeledFrame.instances_to_show` property.

The logic is that it returns all user instances plus all predicted instances that are "unused" ([here](https://github.com/murthylab/sleap/blob/a2738ec10ee5dd2c16df678edfddc1d0eb8aadc8/sleap/instance.py#L1544)).

This is meant to address the case where we have predicted instances that we double-clicked to create a user instance from. In that case, we'd want to not display those since the user instances will be overlapping with them.

The issue is that our logic for finding the "unused" predicted instances is a little weird. It's defined in the `LabeledFrame.unused_predictions` property, in which there's a note about the approach we use ([here](https://github.com/murthylab/sleap/blob/a2738ec10ee5dd2c16df678edfddc1d0eb8aadc8/sleap/instance.py#L1520)).

The weirdness comes from the fact that we use the `PredictedInstance.track` to both determine whether it's "unused" and also to sort the order they're returned in, even though the presence of a track has no bearing on whether the user has double-clicked it or not. I think this was a kludgy fix to address #518, but some of the logic might pre-date even that.

Given this mess, we may want to change how all of this works in the future, so it might be a good idea to avoid using if possible (or at least dropping TODO comment mentioning that it might need to be revisited in the future).

sleap/gui/widgets/video.py
```diff
@@ -485,6 +485,8 @@ def showInstances(self, show):
         """
         for inst in self.instances:
             inst.showInstances(show)
+        for inst in self.predicted_instances:
+            inst.showInstances(show)
```
talmo, line 489: Test coverage for this line might be good given the mess related to `LabeledFrame.instances_to_show` mentioned above.

Also, in light of #518, perhaps the tests should check that the color of the `QtInstance` objects don't change? I think the colors are determined when they're instantiated, so toggling the visibility might not affect it, but it would be good to check.

If the test for color is unnecessary after looking into it, that's fine.

talmo (changes requested):

sleap/gui/app.py
roomrys: Updated this to one again check if `self.state['labeled_frame'] is None`, but still use `LabeledFrame.instances_to_show`. However, I added a note to revisit this line if either the code for `LabeledFrame.instances_to_show` or `LabeledFrame.unused_predictions` changes.

sleap/gui/app.py
roomrys: The nit was well-taken. Lets not rely on implicit casting. This line has been changed to the suggested `n_instances > 0`.

sleap/gui/app.py
roomrys: Updated this to one again check if self.state['labeled_frame'] is None, but still use LabeledFrame.instances_to_show. However, I added a note to revisit this line if either the code for LabeledFrame.instances_to_show or LabeledFrame.unused_predictions changes.

sleap/gui/widgets/video.py
roomrys: Test coverage has been added, but is skipped for ubuntu.

talmo (approved):

Issue #695, opened 2022-04-02
Many suggested frames slow down GUI


olinesn:
Hello,

Writing to make a note that having a long list of suggested frames can substantially slow down the gui. 

Sometimes I'll scrub through a video and add current frame when I get to tricky frames, or regions where I'll be doing a lot of tracking. I then run inference on this large list of suggested frames, sort by Mean Score, and label lots of the worst-scoring frames. However, the gui gets really laggy. I know that it's specifically the list of labelling suggestions, because after I run inference on the list, if I "Clear all" the problem goes away, even if I keep the predicted instances.

Best,
Stefan

olinesn:
To be specific, one example would be moving to a fresh frame, and then double-clicking on a predicted node to begin correcting. There's a lag between the double-click and when the node becomes editable. A second example is that scrubbing through frames by holding down the GO TO Next Frame button becomes laggy with hundreds of suggested frames.

roomrys:
[SLEAP v1.2.3](https://sleap.ai/installation.html) has been released and includes speed-up when searching labeling suggestions.

PR #709, opened 2022-04-18
Speed-up labeling suggestions look-up

roomrys:
### Description
Use the data cache to look-up labeling suggestions.

### Types of changes

- [ ] Bugfix
- [ ] New feature
- [ ] Build / CI changes
- [ ] Documentation Update
- [x] Other (Speed-Up)

Many suggested frames slow down GUI #695

- [x] Review the [guidelines for contributing](https://github.com/murthylab/sleap/wiki/Developer-Guide) to this repository
- [ ] Read and sign the [CLA](https://github.com/murthylab/sleap/blob/develop/sleap-cla.pdf) and add yourself to the [authors list](https://github.com/murthylab/sleap/blob/develop/AUTHORS)

:heart:

### DIFF
--- a/sleap/gui/app.py
+++ b/sleap/gui/app.py
@@ -65,7 +65,7 @@ from PySide2.QtWidgets import QMessageBox
 import sleap
 from sleap.gui.dialogs.metrics import MetricsTableDialog
 from sleap.skeleton import Skeleton
-from sleap.instance import Instance
+from sleap.instance import Instance, LabeledFrame
 from sleap.io.dataset import Labels
 from sleap.info.summary import StatisticSeries
 from sleap.gui.commands import CommandContext, UpdateTopic
@@ -265,7 +265,7 @@ class MainWindow(QMainWindow):
             self.commands.showImportVideos(filenames=filenames)
 
     @property
-    def labels(self):
+    def labels(self) -> Labels:
         return self.state["labels"]
 
     @labels.setter
@@ -1280,7 +1280,9 @@ class MainWindow(QMainWindow):
             if suggestion_list:
                 labeled_count = 0
                 for suggestion in suggestion_list:
-                    lf = self.labels.get((suggestion.video, suggestion.frame_idx))
+                    # TODO(LM): use labels.get after allowing kwargs to __getitem__ and get
+                    lf = self.labels.find(video=suggestion.video, frame_idx=suggestion.frame_idx)
+                    lf = None if len(lf) == 0 else lf[0]
                     if lf is not None and lf.has_user_instances:
                         labeled_count += 1
                 prc = (labeled_count / len(suggestion_list)) * 100
--- a/sleap/gui/app.py
+++ b/sleap/gui/app.py
@@ -1280,9 +1280,7 @@ class MainWindow(QMainWindow):
             if suggestion_list:
                 labeled_count = 0
                 for suggestion in suggestion_list:
-                    # TODO(LM): use labels.get after allowing kwargs to __getitem__ and get
-                    lf = self.labels.find(video=suggestion.video, frame_idx=suggestion.frame_idx)
-                    lf = None if len(lf) == 0 else lf[0]
+                    lf = self.labels.get((suggestion.video, suggestion.frame_idx), use_cache=True)
                     if lf is not None and lf.has_user_instances:
                         labeled_count += 1
                 prc = (labeled_count / len(suggestion_list)) * 100
--- a/sleap/gui/app.py
+++ b/sleap/gui/app.py
@@ -1280,7 +1280,9 @@ class MainWindow(QMainWindow):
             if suggestion_list:
                 labeled_count = 0
                 for suggestion in suggestion_list:
-                    lf = self.labels.get((suggestion.video, suggestion.frame_idx), use_cache=True)
+                    lf = self.labels.get(
+                        (suggestion.video, suggestion.frame_idx), use_cache=True
+                    )
                     if lf is not None and lf.has_user_instances:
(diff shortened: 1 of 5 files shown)


sleap/gui/app.py
```diff
@@ -1280,7 +1280,9 @@ def _has_topic(topic_list):
             if suggestion_list:
                 labeled_count = 0
                 for suggestion in suggestion_list:
-                    lf = self.labels.get((suggestion.video, suggestion.frame_idx))
+                    lf = self.labels.get(
+                        (suggestion.video, suggestion.frame_idx), use_cache=True
+                    )
```
talmo, line 1285: Add coverage for this?

sleap/gui/dataviews.py
```diff
@@ -473,7 +473,7 @@ def item_to_data(self, obj, item):
         item_dict["frame"] = int(item.frame_idx) + 1  # start at frame 1 rather than 0
 
         # show how many labeled instances are in this frame
-        lf = labels.get((item.video, item.frame_idx))
+        lf = labels.get((item.video, item.frame_idx), use_cache=True)
```
talmo, line 476: Coverage

sleap/io/dataset.py
```diff
@@ -629,7 +629,9 @@ def __contains__(self, item) -> bool:
                 return self.find_first(item[0], item[1].tolist()) is not None
         raise ValueError("Item is not an object type contained in labels.")
 
-    def __getitem__(self, key, *args) -> Union[LabeledFrame, List[LabeledFrame]]:
+    def __getitem__(
+        self, key, *args, **kwargs
+    ) -> Union[LabeledFrame, List[LabeledFrame]]:
```
talmo, line 634: [`__getitem__`](https://docs.python.org/3.7/reference/datamodel.html#object.__getitem__) is syntactic sugar for `[]` indexing -- this does not support kwargs and we shouldn't encourage direct usage of this method.

I think if anything, maybe we should move the logic to the `get()` method and just use `__getitem__` as a wrapper for that.

sleap/io/dataset.py
```diff
@@ -689,14 +691,14 @@ def __getitem__(self, key, *args) -> Union[LabeledFrame, List[LabeledFrame]]:
         else:
             raise KeyError("Invalid label indexing arguments.")
 
-    def get(self, *args) -> Union[LabeledFrame, List[LabeledFrame]]:
+    def get(self, *args, **kwargs) -> Union[LabeledFrame, List[LabeledFrame]]:
```
talmo, line 694: See comments above re: `__getitem__`.

Moving most of the contents of `__getitem__` to this `get()` method would also be a good opportunity to properly typehint here since `*args, **kwargs` make it really difficult to know how to call methods and should be avoided where possible.

talmo (changes requested):

sleap/gui/app.py
roomrys: I added a test in `test_dataset.py` called `test_suggestions_cache` that ensures the `labels.get((video, index), use_cache=True)` works as expected, but I can create a new test in `test_app.py` that runs through this line specifically.

sleap/io/dataset.py
```diff
@@ -629,74 +629,137 @@ def __contains__(self, item) -> bool:
...
+            slice,
+            np.integer,
+            np.ndarray,
+            list,
+            range,
+        ],
+    ) -> Union[LabeledFrame, List[LabeledFrame]]:
+        """Return labeled frames matching key or return `None` if not found.
+
+        This makes `labels[...]` save and will not raise an exception if the
```
talmo, line 655: Typo: save -> safe?

sleap/io/dataset.py
```diff
@@ -903,7 +966,7 @@ def frames(self, video: Video, from_frame_idx: int = -1, reverse=False):
             yield self._cache._frame_idx_map[video][idx]
 
     def find_first(
-        self, video: Video, frame_idx: Optional[int] = None
+        self, video: Video, frame_idx: Optional[int] = None, use_cache: bool = False
```
talmo, line 969: Add `use_cache` to the docstring

sleap/io/dataset.py
```diff
@@ -629,74 +629,137 @@ def __contains__(self, item) -> bool:
...
+            else:
+                raise KeyError("Invalid label indexing arguments.")
 
-        This is a safe version of `labels[...]` that will not raise an exception if the
-        item is not found.
-        """
-        try:
-            return self.__getitem__(*args)
         except KeyError:
             return None
```
talmo, line 764: Maybe I'm misreading this but it looks like there's no way to actually have those errors raised, even when calling from `__getitem__` since it'll get caught here and return `None`.

Do we maybe want to have a kwarg in this method that toggles re-raising the exceptions when getting to this line? Something like:

```
def get(self, key, *secondary_key, use_cache=False, raise_error=False):
    try:
        ...
    except KeyError as e:
        if raise_error:
            raise e
        else:
            return None
```


Then we can call it from `__getitem__` with:
```
self.get(key, *secondary_key, raise_error=True)
```

talmo (changes requested):

sleap/io/dataset.py
roomrys: You are correct, any errors raised now just return `None`. I was unsure if we actually wanted errors raised or if we preferred returning `None`. I can add a `kwarg` for raising errors.

sleap/io/dataset.py
roomrys: Yes, typo

talmo (approved):

Issue #713, opened 2022-04-27
Loading new skeleton from file causes labels to contain multiple conflicting skeletons


e-velin:
Hi, 

I'm trying to merge two sleap projects. In both of the projects I used the same skeleton but in the first project initially I annotated the head as well, later I decided I don't need that and deleted it. Now I can see that the skeleton file still contains the head but on the annotated frames they are not visible. When I'm trying to merge the two projects, I get the error message 'ValueError: Labels.skeleton can only be used when there is only a single skeleton saved in the labels. Use Labels.skeletons instead.' So I tried to load in into the first project a clean skeleton json file without that does not contain the head (the skeleton file that I saved from the second project). The skeleton number doesn't change in the GUI to the number of the loaded skeleton, but if I save the skeleton file into a new json file, the new, loaded skeleton number will appear. Yet, I get the same error message when I try to merge the two projects. However, when I close the first project and reopen it, it still contains the old skeleton file with the head node, even when I save the skeleton into a json file. Can you help with this please? This happens both is v1.1.5 and v1.2.2. 

Similar to this: https://github.com/murthylab/sleap/issues/592

roomrys:
Hi @e-velin,

No code-based fix yet, but an update as to why the problem may be occurring. 

I found that if I follow the steps above, but refrain from loading a new skeleton into the project (namely **Project 1** steps 4-6), then I get no error messages. **It seems the error is occurring because there is more than one skeleton in a single project.** In the case that the skeletons are different in the two projects, then I get the following warning upon "Merge into Project":

![image](https://user-images.githubusercontent.com/38435167/165601463-b692edb4-9e96-4650-9ad2-e53dfb5e3104.png)

but, the merge still completes successfully. This, it does not matter if the skeletons in the two projects are different (as the skeletons will be merged), but it does matter if a single project has more than one skeleton - see the error in previous comment.

I will keep you updated,
Liezl

cosimogonnelli:
Hi @roomrys,

I am getting the same error as above. In my case is happening when I try to add to my main project a `predictions.slp' file.
I tried to: open a new project, load the same skeleton as my main project, and merge the predictions. This process works but now I have a new project made of predictions and my original project. I tried then to merge the new project into my main project but I still get the same error. 

What I am try to do is to merge into my main project the prediction.slp file generated by using sleap-track, to be able to expand my dataset.

Could somebody help with this issue? 
Please let me know if you need more info.

Thanks 

[...]

PR #730, opened 2022-05-11
Replace skeleton in project when Load Skeleton (GUI)

roomrys:
### Description
Remove ability to add multiple skeletons to project through "Load Skeleton" on gui.

### Types of changes

- [x] Bugfix
- [ ] New feature
- [ ] Build / CI changes
- [ ] Documentation Update
- [ ] Other (explain)

Skeleton incompatibility #713

- [x] Review the [guidelines for contributing](https://github.com/talmolab/sleap/wiki/Developer-Guide) to this repository

:heart:

### DIFF
--- a/sleap/gui/commands.py
+++ b/sleap/gui/commands.py
@@ -1527,7 +1527,7 @@ class OpenSkeleton(EditCommand):
                 context.state["skeleton"] = sk_list[0]
 
         if context.state["skeleton"] not in context.labels:
-            context.labels.skeletons.append(context.state["skeleton"])
+            context.labels.skeletons = [context.state["skeleton"]]
 
 
 class SaveSkeleton(AppCommand):
--- a/tests/gui/test_commands.py
+++ b/tests/gui/test_commands.py
@@ -76,7 +76,7 @@ def test_open_skeleton(min_labels):
 
     # Create command context and params. Load skeleton.
     commands: CommandContext = CommandContext.from_labels(labels)
-    skeleton_filename = "tests\\data\\skeleton\\fly_skeleton_legs.json"
+    skeleton_filename = "tests/data/skeleton/fly_skeleton_legs.json"
     params: dict = {"filename": skeleton_filename}
 
     # Check that new skeleton replaced skeleton instead of appending to skeletons list.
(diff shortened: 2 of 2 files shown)


talmo (changes requested):
This is ok, but it can also lead to some strange behavior if there are existing instances with a different skeleton.

For example, I'm assuming this will fail:
```py
for instance in labels.instances():
    assert instance.skeleton == labels.skeleton
```

Replacing the skeleton on each instance will work, but what happens to the nodes and points already stored on them?

What about the global [`Labels.nodes`](https://github.com/talmolab/sleap/blob/f9702b0eb2689c24873a4a4a04830056c1272ea6/sleap/io/dataset.py#L415)? This is meant to keep track of all the nodes for serialization/deserialization. This is important if we had multiple skeletons, some of which might share the same nodes.

At a minimum, let's add a test for saving → loading and checking that the skeleton matches.

roomrys:
@talmo the `OpenSkeleton` command has the line `topics = [UpdateTopic.skeleton]` ([commands.py: 1504](https://github.com/talmolab/sleap/blob/f9702b0eb2689c24873a4a4a04830056c1272ea6/sleap/gui/commands.py#L1504)) which then results in a callback in [app.py L1265](https://github.com/talmolab/sleap/blob/develop/sleap/gui/app.py#L1265) which handles the updating of nodes and edges. 
```
        if _has_topic([UpdateTopic.skeleton]):
            self.skeletonNodesTable.model().items = self.state["skeleton"]
            self.skeletonEdgesTable.model().items = self.state["skeleton"]
            self.skeletonEdgesSrc.model().skeleton = self.state["skeleton"]
            self.skeletonEdgesDst.model().skeleton = self.state["skeleton"]

            if self.labels.skeletons:
                self.suggestions_form_widget.set_field_options(
                    "node", self.labels.skeletons[0].node_names
                )
```
I will add a test to verify what I think is happening is actually happening.

roomrys:
Update: the call back does not update the skeletons in the instances nor the global nodes list `Label.nodes`. I will have to add some functionality similar to what happens when we Merge Projects (and skeletons) when we call `OpenSkeleton`.

Issue #729, opened 2022-05-11
Exporting analysis HDF5 files for multi-video projects


ShivanshDave:
This is what I think is happening: Upon "File > Export Analysis H5..." SLEAP successfully saves the h5 file. No errors in terminal. But the issue is that SLEAP overwrites predictions from all videos on the same matrix (ie '/tracks') even though there's only one video open (and selected by default) in SLEAP GUI. -- Thus, h5 data is completely unusable, because it overwrites them as opposed to create different matrix for each video. (I think this might be the reason behind issue #611 that showed same frame-counts) 

Perhaps there is no error, and I am reading h5 file wrong. If so, please suggest how to access predictions from individual videos in the HDF5 file.  -- I tried using SLEAP 1.2.x on Windows10 to export, and reading hdf5 file in Matlab using the following command:
`tracks_matrix = h5read(h5file, '/tracks')`
`X_value = tracks_matrix(frame_ind, point_ind, 1)`
`Y_value = tracks_matrix(frame_ind, point_ind, 2)`

Ideally, it would be great to have two options in File menu, to export analysis from either the current video or all videos in the project.

talmo:
Hi @ShivanshDave,

Yep, you're not doing anything wrong -- we generally keep the labeling project separate from the tracking, so the idea is that you generate individual `.slp` files for each session/video after you have a trained model that you're happy with. Consequently, the analysis HDF5 file is intended for single videos.

A convenient way of following this workflow is to use the [`sleap-track`](https://sleap.ai/guides/cli.html#sleap-track) CLI to run the tracking in batch for each video and create predictions files for each one.

We're planning on adding a new GUI element to make this a bit easier to work with (prototype in #494), but in the meantime one thing you can do is to automate this via a script like in #668.

@roomrys: I think we can simplify this into a smaller enhancement for the short term by just adding a menu (+CLI) option to export multiple HDF5 analysis files for multi-video SLP files. Even easier might be to just add this functionality to the existing Export Analysis HDF5 command with no interactivity for now.

roomrys:
Hi @ShivanshDave ,

The new release of SLEAP v1.2.4 is now available for [installation](https://sleap.ai/installation.html) and includes this fix.

Thanks,
Liezl

PR #768, opened 2022-05-27
Create multiple analysis files for multi-video projects

roomrys:
### Description
Add a menu (+CLI) option to export multiple HDF5 analysis files for multi-video SLP files.

### Types of changes

- [ ] Bugfix
- [x] New feature
- [ ] Build / CI changes
- [ ] Documentation Update
- [ ] Other (explain)

Exporting analysis HDF5 files for multi-video projects[#729](https://github.com/talmolab/sleap/issues/729)

:heart:

### DIFF
--- a/sleap/gui/commands.py
+++ b/sleap/gui/commands.py
@@ -293,9 +293,9 @@ class CommandContext:
         """Show gui to save project as a new file."""
         self.execute(SaveProjectAs)
 
-    def exportAnalysisFile(self):
+    def exportAnalysisFile(self, all_videos: bool = False):
         """Shows gui for exporting analysis h5 file."""
-        self.execute(ExportAnalysisFile)
+        self.execute(ExportAnalysisFile, all_videos=all_videos)
 
     def exportLabeledClip(self):
         """Shows gui for exporting clip with visual annotations."""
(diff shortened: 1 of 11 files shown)


tests/io/test_convert.py
```diff
@@ -0,0 +1,64 @@
...
+    output_prefix = slp_path.with_suffix("")
+    args = f"--format analysis {slp_path}".split()
+    sleap_convert(args)
+    assert_analysis_existance(output_prefix)
+
+    # Specify output and retest
+    output_prefix = tmpdir.with_name("prefix")
+    args = f"--format analysis -o {output_prefix} {slp_path}".split()
+    sleap_convert(args)
+    assert_analysis_existance(output_prefix)
```
roomrys, line 41: lets make this into a `call_and_assert_with_output()` function

roomrys (commented):
One "problem", a few improvements.

sleap/gui/commands.py
```diff
@@ -956,25 +953,53 @@ class ExportAnalysisFile(AppCommand):
...
 
-        filename, selected_filter = FileDialog.save(
-            context.app,
-            caption="Export Analysis File...",
-            dir=default_name,
-            filter="SLEAP Analysis HDF5 (*.h5)",
-        )
+        output_paths = []
+        analysis_videos = []
+        for video in videos:
```
talmo, line 984: Not a fan of doing this in a loop here. It will pop up a dialog for every single video in the project -- some of which could have 30+.

If there is more than one video, this command should instead show a `FileDialog.openDir` to prompt the user to pick a directory to save everything to. They lose the ability to customize the filename, but that's way better than clicking "Save" 30+ times.

talmo (changes requested):
One more change needed for UX, but otherwise LGTM

sleap/gui/commands.py
roomrys: Agreed - this could be very annoying. I changed this to open a `FileDialog.openDir` when multiple videos exist, otherwise use `FileDialog.save` for single video projects.

sleap/gui/commands.py
```diff
@@ -956,25 +953,53 @@ class ExportAnalysisFile(AppCommand):
...
+            all_videos = context.labels.videos
+        else:
+            all_videos = [context.state["video"] or context.labels.videos[0]]
+
+        # Check for labeled frames in each video
+        videos = []
+        for video in all_videos:
+            lfs = labels.get(video)
+            if len(lfs) != 0:
+                videos.append(video)
```
roomrys, line 986: simplify with `videos = [video for video in all_videos if len(labels.get(video)) != 0]`

sleap/gui/commands.py
```diff
@@ -956,25 +954,76 @@ class ExportAnalysisFile(AppCommand):
...
+            # Allow user to specify the filename
+            use_default = False
+            dirname = str(fn.parent)
+        else:
+            # Allow user to specify directory, but use default filenames
+            use_default = True
+            dirname = FileDialog.openDir(
+                context.app,
+                caption="Select Folder to Export Analysis Files...",
+                dir=str(PurePath(default_name).parent),
```
roomrys, line 1003: Simplify with `dir=str(fn.parent)`

roomrys (commented):
A few small suggestions for improvement.

tests/gui/test_commands.py
```diff
@@ -63,3 +69,154 @@ def test_get_new_version_filename():
...
+    shutil.copy(small_robot_mp4_vid.backend.filename, tmpdir / "session2" / "video.mp4")
+
+    labels.videos[0].backend.filename = str(tmpdir / "session1" / "video.mp4")
+    labels.videos[1].backend.filename = str(tmpdir / "session2" / "video.mp4")
+
+    params = {"all_videos": True}
+    okay = ExportAnalysisFile_ask(context=context, params=params)
+    assert okay == True
+    ExportAnalysisFile.do_action(context=context, params=params)
+    assert_videos_written(num_videos=2)
```
talmo, line 213: One more edge case to handle: paths are different, but video filename is the same. This is a common scenario where we have different folder names associated with each video (e.g., session timestamp) but the filename itself is the same.

In this case, the current implementation will overwrite the analysis files. There's a design decision to be made here, but I think the easiest thing would be to just name the output files by their video index within the project.

 See also notes on the metadata that should be added in `write_tracking_h5.py`

sleap/info/write_tracking_h5.py
```diff
@@ -8,10 +7,15 @@
...
-* "node_names"         shape: nodes
+* "track_occupancy"    (shape: tracks * frames)
+* "tracks"             (shape: frames * nodes * 2 * tracks)
+* "track_names"        (shape: tracks)
+* "node_names"         (shape: nodes)
+* "edge_names"         (shape: nodes - 1)
+* "edge_inds"          (shape: nodes - 1)
+* "point_scores"       (shape: frames * nodes * tracks)
+* "instance_scores"    (shape: frames * tracks)
+* "tracking_scores"    (shape: frames * tracks)
```
talmo, line 18: Also include:

- `"labels_path"`: Path to the source .slp file (if available from GUI context)
- `"video_path"`: Path to source video (`video.backend.filename`)
- `"video_ind"`: Scalar integer index of the video within the source labels

Also nice to have:
- `"provenance"`: Copied from `labels.provenance`. Might be easier to convert to json string first.

talmo (changes requested):
Important edge case to be considered -- also exposes the need for more metadata.

tests/gui/test_commands.py
roomrys: I believe the naming should be some-what intuitive/memorable. How about we compromise with a name of `<project_prefix>.<video_idx>_<video_name>.analysis.h5`?

talmo (approved):
LGTM!

PR #787, opened 2022-06-10
Allow user to set grayscale when replacing videos (mp4/avi only) 

roomrys:
### Description
Show the same dialog for adding videos upon replacing videos (`ImportItemWidget`) which gives users the ability to select grayscale on mp4/avi videos -  no other import-able backends currently support grayscale.

The `ReplaceVideo` command continues to only work properly if the replacement video has the same extension as the current video - this limitation was not addressed in this PR.

### Types of changes

- [ ] Bugfix
- [x] New feature
- [ ] Build / CI changes
- [ ] Documentation Update
- [ ] Other (explain)

None.

:heart:


sleap/gui/commands.py
```diff
@@ -1488,30 +1489,106 @@ def do_action(context: CommandContext, params: dict):
...
+                    )
+                    query = QueryDialog(
+                        "Replace Video: Truncate Labeled Frames", message, context.app
+                    )
+                    query.accepted.connect(lambda: _trim_labeled_frames(lfs))
+                    # FIXME: Ideally we do not mess with backend before approval
+                    query.rejected.connect(
+                        lambda: _reset_video_backend(video, old_path, old_grayscale)
+                    )
+                    query.exec_()
```
roomrys, line 1586: This bit of code will pop up for EVERY video that has labeled frames past the replacement video length - is it safe to assume that there will not be many cases where the above is true? How often do our users replace videos? How many videos do they replace at once? Otherwise, we will need to create a pop-up box similar to the `ImportItemWidget` which shows all problem videos in a single dialog.

sleap/gui/commands.py
```diff
@@ -1488,30 +1489,106 @@ def do_action(context: CommandContext, params: dict):
...
+            old_path = video.backend.filename
+            old_grayscale = video.backend.grayscale
+            new_path = import_params["filename"]
+
+            # TODO: Will need to create a new backend if import has different extension.
+            # Currently reset is only functional for MediaVideo backend.
+            # FIXME: Ideally we would only mess with the backend after getting approval
+            # ...but, use this to read in the number of frames. Could use cv2 here....
+            if new_path != video.backend.filename:
+                _reset_video_backend(video, new_path, import_params)
```
roomrys, line 1556: We really shouldn't be changing the backend before getting approval from the user. To separate into `ask` and `do_action`, we could read the number of frames using cv2:
```py
reader = cv2.VideoCapture(new_path)
last_vid_frame = int(reader.get(cv2.CAP_PROP_FRAME_COUNT))
```

roomrys (commented):
Ask permission before changing backend (even if it will be reset later on).

sleap/gui/commands.py
roomrys: The warning is now built into the `ImportItemWidget` that pops up to create our list of files to import.

sleap/gui/commands.py
roomrys: I chose to go with the above recommendation and create a new reader for each video to (possibly) import.

sleap/gui/commands.py
```diff
@@ -1516,30 +1517,96 @@ def do_action(context: CommandContext, params: dict):
...
 
-        okay = MissingFilesDialog(filenames=paths, replace=True).exec_()
+        def _get_truncation_message(truncation_messages, path, video):
+            reader = cv2.VideoCapture(path)
+            last_vid_frame = int(reader.get(cv2.CAP_PROP_FRAME_COUNT))
+            lfs: List[LabeledFrame] = list(context.labels.get(video))
+            if lfs is not None:
+                lfs.sort(key=lambda lf: lf.frame_idx)
+                last_lf_frame = lfs[-1].frame_idx
+                lfs = [lf for lf in lfs if lf.frame_idx > last_vid_frame]
```
roomrys, line 1554: This line is only used for displaying information to user. Do we need it (i.e., is it worth any lag that it causes)?

sleap/gui/commands.py
```diff
@@ -1516,30 +1517,96 @@ def do_action(context: CommandContext, params: dict):
...
+                    message = (
+                        "<p><strong>Warning:</strong> Replacing this video will "
+                        f"remove {len(lfs)} labeled frames.</p>"
+                        f"<p><em>Current video</em>: <b>{Path(video.filename).name}</b>"
+                        f" (last label at frame {last_lf_frame})<br>"
+                        f"<em>Replacement video</em>: <b>{Path(path).name}"
+                        f"</b> ({last_vid_frame} frames)</p>"
+                    )
+                    # Assumes that a project won't import the same video multiple times
+                    truncation_messages[path] = message
```
roomrys, line 1567: Will a user ever import the same video more than once?

sleap/gui/commands.py
```diff
@@ -1516,30 +1517,96 @@ def do_action(context: CommandContext, params: dict):
...
+
+            return truncation_messages
+
+        # Warn user: newly added labels will be discarded if project is not saved
+        if not context.state["filename"] or context.state["has_changes"]:
+            QMessageBox(
+                text=("You have unsaved changes. Please save before replacing videos.")
+            ).exec_()
+            return False
```
roomrys, line 1576: This is triggered even when labels are not added/changed (but another "change" was registered). Is there a way to check that just the labels object has changes?

roomrys (commented):
Added a few questions/considerations for my reviewer.

sleap/gui/commands.py
talmo: This is pretty minimal -- I'm ok with it in favor of being more verbose about potential catastrophic situations

sleap/gui/commands.py
talmo: They shouldn't be able to!

sleap/gui/commands.py
talmo: That would be amazing -- I've looked into it in the past but it doesn't seem trivial. Search for `"has_changes"` across the codebase.

In general, it's just a tricky pattern to implement. Here's some relevant literature: https://refactoring.guru/design-patterns/state

sleap/gui/dialogs/importvideos.py
```diff
@@ -152,7 +155,10 @@ def __init__(self, filenames: List[str], *args, **kwargs):
                             this_type = import_type
                             break
                 if this_type is not None:
-                    import_item_widget = ImportItemWidget(file_name, this_type)
+                    message = messages[file_name] if file_name in messages else str()
```
talmo, line 158: Nit: `... else ""` instead of `... else str()`

sleap/gui/dialogs/importvideos.py
```diff
@@ -525,6 +538,26 @@ def paint(self, painter, option, widget=None):
...
+        self.layout.addWidget(self.message)
+        self.setLayout(self.layout)
+
+    def boundingRect(self) -> QRectF:
+        """Method required by Qt."""
+        return QRectF()
+
+    def paint(self, painter, option, widget=None):
+        """Method required by Qt."""
+        pass
```
talmo, line 558: Really?? I would think you wouldn't need to override it, or at least have a `super().paint(...)` in there?

talmo (changes requested):

sleap/gui/commands.py
roomrys: @talmo I was able to. We can open an issue about this and fix it in another PR?

sleap/gui/dialogs/importvideos.py
roomrys: Admittedly, I meant to test this myself and had not gotten to it - will test before next review.

sleap/gui/commands.py
```diff
@@ -1516,30 +1517,96 @@ def do_action(context: CommandContext, params: dict):
...
+        import_list = params["import_list"]
 
-        for video, new_path in zip(context.labels.videos, new_paths):
-            if new_path != video.backend.filename:
-                video.backend.filename = new_path
-                video.backend.reset()
+        for import_item, video in import_list:
+            import_params = import_item["params"]
+
+            # TODO: Will need to create a new backend if import has different extension.
```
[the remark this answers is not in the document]
roomrys: I edited `MissingFilesDialog` to only allow the current video type using the `replace` kwarg, but also raise an error in `ReplaceVideo.do_action()`.

sleap/gui/dialogs/importvideos.py
roomrys: It does not seem to be required by Qt.

sleap/gui/dialogs/importvideos.py
```diff
@@ -91,7 +95,7 @@ class ImportParamDialog(QDialog):
         filenames (list): List of files we want to import.
     """
 
-    def __init__(self, filenames: List[str], *args, **kwargs):
+    def __init__(self, filenames: List[str], messages: dict = dict(), *args, **kwargs):
```
talmo, line 98: Big no-no here! Mutable types should not be instantiated as function defaults! More info: https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

See also the [typing module docs](https://docs.python.org/3.7/library/typing.html) for the proper way to typehint `dict`s in python 3.7.

What we want here is something like:
```
    def __init__(self, filenames: List[str], messages: Optional[Dict[str, str]] = None, *args, **kwargs):
        if messages is None:
            messages = dict()
```

sleap/gui/dialogs/importvideos.py
```diff
@@ -269,7 +268,9 @@ class ImportItemWidget(QFrame):
         import_type (dict): Data about user-selectable import parameters.
     """
 
-    def __init__(self, file_path: str, import_type: dict, *args, **kwargs):
+    def __init__(
+        self, file_path: str, import_type: dict, message: str = str(), *args, **kwargs
```
talmo, line 272: Type hint should be `import_type: Dict[str, Any]` or something

talmo (changes requested):
A couple more quick fixes and you're good to go

sleap/gui/dialogs/importvideos.py
roomrys: Got it! Thanks

Issue #826, opened 2022-07-05
Crop size should always be compatible with max stride


isabelperezf:
Hi, @auesro and I we marked and corrected around one thousand of frames  with multianimal-bottom-up. Finally we want to compare different types of backbones. We chose Unet model for the centroid model, because we obtained a good result, and Unet, leap, hourglass, pretrained and resnet backbones in the centered instances. We didn´t have problems with unet and leap when we've trained on Colab, but when we`ve tried to do the training with the rest of them we've obtained different ValueErrors. Can someone help us?
Thanks. 

Pretained
```
INFO:numexpr.utils:NumExpr defaulting to 2 threads.
INFO:sleap.nn.training:Versions:
SLEAP: 1.2.4
TensorFlow: 2.8.2
Numpy: 1.21.5
Python: 3.7.13
OS: Linux-5.4.188+-x86_64-with-Ubuntu-18.04-bionic
INFO:sleap.nn.training:Training labels file: proyectoprueba_v9.pkg.slp
INFO:sleap.nn.training:Training profile: centered_instance.json
INFO:sleap.nn.training:
INFO:sleap.nn.training:Arguments:
INFO:sleap.nn.training:{
    "training_job_path": "centered_instance.json",
    "labels_path": "proyectoprueba_v9.pkg.slp",
    "video_paths": [
        ""
    ],
    "val_labels": null,
    "test_labels": null,
    "tensorboard": false,
    "save_viz": false,
    "zmq": false,
    "run_name": "",
    "prefix": "",
    "suffix": "",
    "cpu": false,
    "first_gpu": false,
    "last_gpu": false,
    "gpu": 0
}
INFO:sleap.nn.training:
INFO:sleap.nn.training:Training job:
INFO:sleap.nn.training:{
    "data": {
        "labels": {
            "training_labels": null,
            "validation_labels": null,
            "validation_fraction": 0.1,
            "test_labels": null,
            "split_by_inds": false,
            "training_inds": null,
            "validation_inds": null,
            "test_inds": null,
            "search_path_hints": [],
            "skeletons": []
        },
        "preprocessing": {
            "ensure_rgb": false,
            "ensure_grayscale": false,
            "imagenet_mode": null,
            "input_scaling": 1.0,
            "pad_to_stride": null,
            "resize_and_pad_to_target": true,
            "target_height": null,
            "target_width": null
        },
        "instance_cropping": {
            "center_on_part": null,
            "crop_size": 144,
[...]

PR #841, opened 2022-07-15
Recalculate crop size if user-specified crop size indivisible by max stride

roomrys:
### Description
The crop-size should always be divisible by the max-stride. Prior to this PR, SLEAP allowed users to overwrite automatic calculation of the crop size (which calculates a crop size where `crop size % max stride = 0`). However, this leads to errors later in the training pipeline.

This PR forces automatic calculation if the user-set crop size is indivisible by the max stride. The returned crop size will be greater than or equal to the user-specified crop size.

### Types of changes

- [x] Bugfix
- [ ] New feature
- [ ] Build / CI changes
- [ ] Documentation Update
- [ ] Other (explain)

Crop size should always be compatible with max stride #826

:heart:

### DIFF
--- a/sleap/nn/data/instance_cropping.py
+++ b/sleap/nn/data/instance_cropping.py
@@ -13,7 +13,7 @@ def find_instance_crop_size(
     padding: int = 0,
     maximum_stride: int = 2,
     input_scaling: float = 1.0,
-    crop_size: Optional[int] = None,
+    min_crop_size: Optional[int] = None,
 ) -> int:
     """Compute the size of the largest instance bounding box from labels.
 
@@ -25,7 +25,7 @@ def find_instance_crop_size(
             architecture.
         input_scaling: Float factor indicating the scale of the input images if any
             scaling will be done before cropping.
-        crop_size: The (optional) crop size set by the user. None if not set.
+        min_crop_size: The (optional) crop size set by the user. None if not set.
 
     Returns:
         An integer crop size denoting the length of the side of the bounding boxes that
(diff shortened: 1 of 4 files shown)


sleap/nn/data/instance_cropping.py
```diff
@@ -13,6 +13,7 @@ def find_instance_crop_size(
     padding: int = 0,
     maximum_stride: int = 2,
     input_scaling: float = 1.0,
+    crop_size: Optional[int] = None,
```
talmo, line 16: Should this be more aptly called `min_crop_size`?

sleap/nn/data/instance_cropping.py
```diff
@@ -24,19 +25,24 @@ def find_instance_crop_size(
...
     """
+    crop_size = 0 if crop_size is None else crop_size
+
     max_length = 0.0
     for inst in labels.user_instances:
         pts = inst.points_array
         pts *= input_scaling
         max_length = np.maximum(max_length, np.nanmax(pts[:, 0]) - np.nanmin(pts[:, 0]))
         max_length = np.maximum(max_length, np.nanmax(pts[:, 1]) - np.nanmin(pts[:, 1]))
+        max_length = np.maximum(max_length, crop_size)
```
talmo, line 45: I'm a little concerned about this since we add the padding afterwards.

Scenario:
1. `crop_size` was `None` in the first round of training, and this function computes it and returns `160`.
2. The user labels more and wants to retrain, so by default we pull in the config from the last round, which prepopulates that field with `160`.
3. It gets passed to this function again. Now it's `160` + `padding` = `176`.

Like, should we maybe just check if the `min_crop_size` is divisible already and/or assume it already includes padding?

talmo (approved):
Minor comments, but feel free to merge if they can be safely ignored

sleap/nn/data/instance_cropping.py
roomrys: We now check if `min_crop_size` is divisible and also assume it already includes padding.

PR #870, opened 2022-07-28
Use saved optical flow to use only on adjacent frames

getzze:
## Description
Using tracking with optical flow, similarity between frames is computed with the optical flow between the current frame and the reference frame. This is fine if the 2 frames are close in time, but it works poorly with say 20 time-steps difference, as the optical flow is poorly computed.

Instead, save the instances shifted by optical flow step by step, so the optical flow is always computed between two adjacent frames.

The `save_shifted_instances` argument of `FlowCandidateMaker` could be made accessible from the GUI, so we could switch between the 2 options.

### Types of changes

- [ ] Bugfix
- [x] New feature
- [ ] Build / CI changes
- [ ] Documentation Update
- [ ] Other (explain)

- #891

:heart:


talmo (changes requested):
Hi @getzze,

Thanks for the contribution! This looks like some interesting functionality and I'd be curious to see how it works in practice.

Since this is making a change to a core component within SLEAP, do you mind adding a short test to [`tests/nn/test_tracker_components.py`](https://github.com/getzze/sleap/blob/use_saved_flow/tests/nn/test_tracker_components.py)?

We just want to make sure nothing strange happens with the other components and that this runs as expected.

Thanks!!

Talmo

sleap/nn/tracking.py
```diff
@@ -97,7 +104,7 @@ class FlowCandidateMaker:
     of_window_size: int = 21
     of_max_levels: int = 3
 
-    save_shifted_instances: bool = False
+    save_shifted_instances: bool = True
```
talmo, line 107: We'd also prefer not to change any defaults. In this case, it's possible that saving a significant number of matched instances may affect performance, so maybe we can leave it at `False` by default and still enable the functionality by manually setting when it's instantiated.

getzze: Yes it makes sense, I reverted it

getzze:
Hi,
I made `save_shifted_instance` an option for the cli and the gui.

I looked at making a test but it's not so simple as testing this requires having some images to run OpenCV optical flow function on them. I cannot promise I will have time to look at it during the next month.

talmo (approved):
lgtm

Issue #876, opened 2022-07-29
Unintuitive output naming for sleap-convert


Monnesee:
# Bug description

When using sleap-convert, it doesn't correctly name the output files, and worse, it adds a period in the middle of the name, making it hard to call on later. It also copies on the input filename, which means if you are setting the output as the filename, it'll do it twice.

## Expected behaviour

Expect output filename to be what I specified in sleap-convert
Essentially that the --output flag works the same as this flag in sleap-track

## Actual behaviour

When specifying [value] for sleap-track --output, 
the filename ends up being [value].000filename

This is causing major issues when trying to call the file due to the period in the middle.
I'm sure there are workarounds for this but I can't help but see this as a bug.

## Your personal set up

- OS:
Google Colab
- Version(s):

SLEAP: 1.2.6
TensorFlow: 2.8.2
Numpy: 1.21.5
Python: 3.7.13

- SLEAP installation method (listed [here](https://sleap.ai/installation.html#)):
  - [ ] [Conda from package](https://sleap.ai/installation.html#conda-package)
  - [ ] [Conda from source](https://sleap.ai/installation.html#conda-from-source)
  - [x] [pip package](https://sleap.ai/installation.html#pip-package)
  - [ ] [M1 Macs](https://sleap.ai/installation.html#m1-macs)

</details>

## Screenshots

Notice: the expected filename should be all_h5/ or at the very least all_h5/filename, but its all_h5.000.h5
![image](https://user-images.githubusercontent.com/29939478/181852925-f5a1d243-646f-418d-9ad9-b822b8fe86e5.png)

Notice: the expected filename should be "what" but it is what.000_[input filename].h5
![image](https://user-images.githubusercontent.com/29939478/181853013-487a3675-2425-4a68-b689-89252452814d.png)

## How to reproduce

When using sleap-track, specify any value as the output. it will add ".000" between your value and the filename.

roomrys:
Hi @Monnesee,

Agreed, this is a very unpredictable feature especially when using SLEAP through the CLI. We originally introduced this bug in #768 when adding support to automatically generate analysis files for all videos in the `.slp`; hence the strange naming convention using the user input as an output prefix, the video index, and the video name to create a unique identifier:
https://github.com/talmolab/sleap/blob/ab63c9d3f2c4725c70a231c58656523c2d6313f2/sleap/io/convert.py#L80-L85

[...]

PR #881, opened 2022-08-02
Fix naming convention for sleap-convert

roomrys:
### Description
In #768, we added a feature to handle creating analysis files for multi-video projects; however, we also implemented a semi-automatic naming convention which is unintuitive to users.

This PR gives users control over the output path (as it should have always been) and only starts creating default names if no output path is given for a video in the project. 

Ex. 3 output names given with 4 videos will generate 1 default name. Output names are mapped to videos based on ordering of names given and video index in the project.

### Types of changes

- [x] Bugfix
- [ ] New feature
- [ ] Build / CI changes
- [ ] Documentation Update
- [ ] Other (explain)

sleap-convert names files with 000.filename #876

:heart:

### DIFF
--- a/.github/workflows/website.yml
+++ b/.github/workflows/website.yml
@@ -8,7 +8,7 @@ on:
       # 'main' triggers updates to 'sleap.ai', 'develop' to 'sleap.ai/develop'
       - main
       - develop
-      - liezl/convert-output-path
+      - liezl/convert_output-path
     paths:
       - "docs/**"
       - "README.rst"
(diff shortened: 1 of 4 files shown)


talmo (approved):
lgtm

one small suggestion: in the docs `<input path>` could maybe be `<slp path>` so it's clear what we mean by input here? Also an example could be good, like:

```
                        Path to output file (optional). The analysis format expects an
                        output path per video in the project. Otherwise, the default
                        naming convention
                        <input path>.<video index>_<video filename>.analysis.h5 will be
                        used for every video without a specified output path. Multiple
                        outputs can be specified, each preceded by --output.
                        
                        Example:
                        > sleap-convert predictions.slp
                        predictions.000_video.analysis.h5
                        predictions.001_video.analysis.h5
                        ...
```

Does that look right?

PR #911, opened 2022-08-12
Add GPU Memory Polling Function

jmdelahanty:
### Description
Imports `subprocess` for invoking `nvidia-smi` query that asks for GPU indexes, unused/available memory, and total memory on the card. Returns a dictionary that has GPU index as the key with fraction of GPU available as the value.

### Types of changes

- [ ] Bugfix
- [x] New feature
- [ ] Build / CI changes
- [ ] Documentation Update
- [ ] Other (explain)

Nope!

- [x] Review the [guidelines for contributing](https://github.com/talmolab/sleap/wiki/Developer-Guide) to this repository
- [x] Read and sign the [CLA](https://github.com/talmolab/sleap/blob/develop/sleap-cla.pdf) and add yourself to the [authors list](https://github.com/talmolab/sleap/blob/develop/AUTHORS)
Are tests required for this function and, if so, what should that look like?
Much redacted documentation in the function, down to write docs elsewhere if you'd like!

:heart:


roomrys:
TODO to make this usable:

Add a flag to `sleap-train` called `--gpu auto` that uses this function to select the GPU with lowest memory usage. It would be then called here:
https://github.com/talmolab/sleap/blob/2688d56172eb297a25d33a47a0b4bdc84a5652d9/sleap/nn/training.py#L1932-L1941

Something like:
```py
        if args.first_gpu:
            sleap.nn.system.use_first_gpu()
            logger.info("Using the first GPU for acceleration.")
        elif args.last_gpu:
            sleap.nn.system.use_last_gpu()
            logger.info("Using the last GPU for acceleration.")
        else:
            if args.gpu == "auto":
                gpu_ind = np.argmin(sleap.nn.system.get_gpu_memory().values())
            else:
                gpu_ind = int(args.gpu)
            sleap.nn.system.use_gpu(gpu_ind)
            logger.info(f"Using GPU {args.gpu} for acceleration.")
```

And then a similar setup for inference here:
https://github.com/talmolab/sleap/blob/2688d56172eb297a25d33a47a0b4bdc84a5652d9/sleap/nn/inference.py#L4270-L4279

sleap/nn/system.py
```diff
@@ -183,3 +184,46 @@ def best_logical_device_name() -> Text:
...
+        gpu_id = row.split(",")[0]
+        available_memory = row.split(",")[1]
+        total_memory = row.split(",")[2]
+
+        # Removing megabyte text returned by nvidia-smi
+        available_memory = available_memory.split()[0]
+        total_memory = total_memory.split()[0]
+
+        # Append percent of GPU available to GPU ID
+        memory_dict[gpu_id] = round(int(available_memory) / int(total_memory), 4)
```
roomrys, line 227: Do we want this to be a percentage?

roomrys:
Although we cannot test gpu features through github actions yet, I tested the new additions locally - they passed.

sleap/nn/system.py
jmdelahanty: I don't know! I wasn't sure if having it as a percentage would be helpful/more readable/understandable so I just left it that way.

jmdelahanty: I also realize now that the comment says percentage but it's not a percent value that's given, so we can change the comment or the value. I have no preference and also don't know what best practice is lol

jmdelahanty:
Neato!

sleap/nn/system.py
jmdelahanty: I think as it is currently written it is a fraction no? Ratio of:

Available Memory
________________
Total Memory

jmdelahanty:
This is so much cleaner! I don't know why I didn't think of using the list indices as the index of the card. That makes a lot more sense. Nice one Liezl!

PR #981, opened 2022-10-05
Added upper limit of the instance count in prediction score labeling suggestion method

hsingchien:
### Description
PR 2/3 for https://github.com/talmolab/sleap/discussions/977
Renamed original instance num input to `instance_limit_lower`. Added `instance_limit_upper` to the method. Updated suggestion config file accordingly. User can now have finer control over the number of instances in generating prediction score based labeling suggestion. 

---

I found it useful to quickly generate frame list with only certain number of visible predicted instances. I am working on a 2-animal project, having 1 instance means failure on the other animal. 
This can be achieved by a small tweak on <prediction_score> suggestion method. I renamed original input <num_instance> to <instance_at_least> added another input <instance_no_more_than>. Now I can just set <predition_score> to 100 and both instance upperbound and lowerbound to be 1, and label suggestion will give back all frames with precisely 1 predicted instance. This obviously enables other frame filtering possibilities as well.

_Originally posted by @hsingchien in https://github.com/talmolab/sleap/discussions/977_

### Types of changes

- [ ] Bugfix
- [x] New feature
- [ ] Build / CI changes
- [ ] Documentation Update
- [ ] Other (explain)

[list open issues here]

- [x] Review the [guidelines for contributing](https://github.com/talmolab/sleap/blob/develop/docs/CONTRIBUTING.md) to this repository

:heart:

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -18,7 +18,6 @@ dependencies:
   # - sleap::pyside2=5.14.1
   - qtpy>=2.0.1
   - conda-forge::pip!=22.0.4
-  - pip
   - pip:
       - "--editable=."
       - "--requirement=./dev_requirements.txt"
--- a/environment.yml
+++ b/environment.yml
@@ -1,4 +1,4 @@
-name: sleap-dev
+name: sleap
 
 dependencies:
   - python=3.7
--- a/environment.yml
+++ b/environment.yml
@@ -1,4 +1,4 @@
-name: sleap
+name: sleap-dev
 
 dependencies:
   - python=3.7
@@ -18,6 +18,7 @@ dependencies:
   # - sleap::pyside2=5.14.1
   - qtpy>=2.0.1
   - conda-forge::pip!=22.0.4
+  - pip
   - pip:
       - "--editable=."
       - "--requirement=./dev_requirements.txt"
--- a/sleap/gui/suggestions.py
+++ b/sleap/gui/suggestions.py
@@ -174,17 +174,19 @@ class VideoFrameSuggestions(object):
         labels: "Labels",
         videos: List[Video],
         score_limit,
-        instance_limit,
+        instance_limit_upper,
+        instance_limit_lower,
         **kwargs,
     ):
         """Method to generate suggestions for proofreading frames with low score."""
         score_limit = float(score_limit)
-        instance_limit = int(instance_limit)
+        instance_limit_upper = int(instance_limit_upper)
+        instance_limit_lower = int(instance_limit_lower)
 
         proposed_suggestions = []
         for video in videos:
             proposed_suggestions.extend(
-                cls._prediction_score_video(video, labels, score_limit, instance_limit)
+                cls._prediction_score_video(video, labels, score_limit, instance_limit_upper, instance_limit_lower)
             )
 
         suggestions = VideoFrameSuggestions.filter_unique_suggestions(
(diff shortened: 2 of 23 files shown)


environment.yml
```diff
@@ -18,6 +18,7 @@ dependencies:
   # - sleap::pyside2=5.14.1
   - qtpy>=2.0.1
   - conda-forge::pip!=22.0.4
+  - pip
```
roomrys, line 21: Did you experience any issues with this removed?

roomrys (changes requested):
A small correction that for an edge case and then we can also add some tests in the tests/gui/test_suggestions.py file. I wrote a sample test that we can add to:
```py
def test_limits_prediction_score(centered_pair_predictions: Labels):
    """Testing suggestion generation using instance limits and prediction score."""
    labels = centered_pair_predictions
    score_limit = 20
    instance_lower_limit = 3
    instance_upper_limit = 3

    # Generate suggestions
    suggestions = VideoFrameSuggestions.suggest(
        labels=labels,
        params={
            "videos": labels.videos,
            "method": "prediction_score",
            "score_limit": score_limit,
            "instance_limit_upper": instance_upper_limit,
            "instance_limit_lower": instance_lower_limit,
        },
    )

    # Confirming every suggested frame meets criteria
    for sugg in suggestions:
        lf = labels.get((sugg.video, sugg.frame_idx))
        pred_instances = [
            inst for inst in lf.instances_to_show if isinstance(inst, PredictedInstance)
        ]
        n_instance_below_score = np.nansum(
            [True for inst in pred_instances if inst.score < score_limit]
        )
        assert n_instance_below_score >= instance_lower_limit
        assert n_instance_below_score <= instance_upper_limit
```
The limits are set score limit is set s.t. we only get 6 suggestions returned (for speed of the assertion loop at the end).

sleap/gui/suggestions.py
```diff
@@ -195,29 +203,35 @@ def prediction_score(
...
+
+            if (
+                n_qualified_instance >= instance_limit_lower
+                and n_qualified_instance <= instance_limit_upper
+            ):
+                # idxs saves qualified frame index at corresponding entry, otherwise the entry is 0
+                idxs[i] = lf.frame_idx
+
+        # Finds non-zero entries in idxs
+        result = sorted(idxs[idxs > 0].tolist())
```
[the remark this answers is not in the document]
hsingchien: Changed accordingly! Thank you!

sleap/gui/suggestions.py
```diff
@@ -195,29 +203,35 @@ def prediction_score(
...
+        labels: "Labels",
+        score_limit: float,
+        instance_limit_upper: int,
+        instance_limit_lower: int,
     ):
         lfs = labels.find(video)
         frames = len(lfs)
-        idxs = np.ndarray((frames), dtype="int")
-        scores = np.full((frames, instance_limit), 100.0, dtype="float")
+        idxs = np.zeros((frames), dtype="int")
```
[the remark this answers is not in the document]
hsingchien: Corrected!

environment.yml
hsingchien: Old yaml gave infinite warnings suggesting some packages require `pip`, without `pip` installation cannot be done. Adding pip solved the problem, so was environment_build.yml.  environment_m1.yml has pip line though.

hsingchien:
I have corrected the comments in suggestions.py . Also added the sample test from @roomrys to confirm all the suggested frames meet the criteria (corrected  `inst.score < score_limit` in the original script to `inst.score <= score_limit`). Also added another test to confirm all the qualified frames in target videos are successfully captured.

tests/gui/test_suggestions.py
```diff
@@ -366,3 +369,62 @@ def test_append_suggestions(small_robot_3_frame_vid: Video, stickman: Skeleton):
...
+        n_instance_below_score = np.nansum(
+            [True for inst in pred_instances if inst.score <= score_limit]
+        )
+        assert n_instance_below_score >= instance_lower_limit
+        assert n_instance_below_score <= instance_upper_limit
+
+    # Confirming all frames meeting the criteria are captured
+    def check_all_predicted_instances(sugg, labels):
+        for video in labels.videos:
+            lfs = labels.find(video)
```
roomrys, line 408: We can remove this outer for loop as there is only a single video in the `centered_pair_predictions` project.
```suggestion
lfs = labels.labeled_frames
```

roomrys (commented):
Minor suggestion to speed-up the tests, then I believe we are good to go.

tests/gui/test_suggestions.py
hsingchien: Done! Also corrected another mistake I made earlier

sleap/gui/suggestions.py
```diff
@@ -195,29 +203,36 @@ def prediction_score(
...
+            # Scores from visible instances in frame
+            pred_fs = lf.instances_to_show
+            frame_scores = np.array(
+                [inst.score for inst in pred_fs if hasattr(inst, "score")]
+            )
+            # Gets the number of instances with scores lower than <score_limit>
+            n_qualified_instance = np.nansum(frame_scores <= score_limit)
+
+            if (
+                n_qualified_instance >= instance_limit_lower
```
talmo, line 228: `instance_limit_lower` has a default of 0, which will evaluate to `True` for empty frames I think. Like:

```py
frame_scores = np.array([])  # == no predicted instances in frame
n_qualified_instance = np.nansum(frame_scores <= score_limit)  # == 0
```

Do we maybe want the lower limit to be 1 or switch to using `>` instead of `>=`?

hsingchien: That's a good point. I guess one of the uses of this feature is to quickly locate the frames where prediction failed, and that would include empty frames (e.g. after a whole video prediction). Therefore, I prefer to keep the option open to users

roomrys (approved):

hsingchien:
Regarding to @talmo 's point earlier, I realized that setting lower limit to 0 will also give out frames with user instances and no predicted instances, which is likely not desirable if user only wants to get frames failed in prediction. A potential solution would be to give user an option to narrow the searching scope to unlabeled frames. This could be a dropdown menu like `target video` that can be applied to all methods or a button/menu command to remove labeled frames from labeling suggestions.

PR #1029, opened 2022-11-04
Shifted instances: make sure ref_instances is not empty

getzze:
### Description
Correction of another bug introduced by `save_shifted_instances=True`. Hopefully the last one...
I had corrected it locally but forgot to include it in the last correction PR #1001.

Loading frames from `self.shifted_instances`, `ref_instances` changes and can become an empty list.
But the test for `ref_instances` being non-empty was before these lines, therefore the following lines crash because `ref_instances` is unexpectedly an empty list.
There are two changes to correct this bug:
- put the code for loading from `self.shifted_instances` above the `len(ref_instances) > 0` test (this is not really needed but it makes the code easier to read).
- `ref_instances` (and `ref_img`) is modified only if it is not empty.

### Types of changes

- [X] Bugfix

### DIFF
--- a/sleap/nn/tracking.py
+++ b/sleap/nn/tracking.py
@@ -150,17 +150,18 @@ class FlowCandidateMaker:
                 matched_item.instances_t,
             )
 
-            if len(ref_instances) > 0:
-                # Check if shifted instance was computed at earlier time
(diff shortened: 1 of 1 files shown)


roomrys (approved):
lgtm

PR #1062, opened 2022-11-28
Tracking: robust assignment of the best score to an instance

getzze:
Another contribution to improve the tracking algorithm :)

This change reduces identity switches (tested on videos with two mice).
The tracking algorithm creates a matrix of similarity of the instances with the existing tracks.
For that, for each track, reference instances from previous time steps are collected in a list and the similarity score between the current instance  and the reference instances are computing, resulting in a list of similarity scores.
Before this patch, the maximum score from this list was computed, to be used as the track-instance similarity entry in the matrix.

The following problem arises in some situations and time step: the similarity scores with anterior reference instances is mainly low, but erroneously high for just one frame (where keypoints were incorrectly located for instance). Because we take the maximum (a non-robust metric), it will results in a high track-instance score and the instance can be assigned incorrectly to this track. And this error propagates.
To mitigate this situation, this patch allows for taking a quantile of the list of scores, which is robust to having outliers. The quantile should be high, 0.95 is the proposed value, but 0.9 would make it even more robust.
The default remains using the maximum score, but can be changed with the CLI option `--tracker.robust 0.95` and in the GUI.

### DIFF
--- a/sleap/config/pipeline_form.yaml
+++ b/sleap/config/pipeline_form.yaml
@@ -271,6 +271,16 @@ inference:
       label: Elapsed Frame Window
       type: int
       default: 5
+    - name: tracking.robust
+      label: 'Robust quantile of similarity scores'
+      help: 'For a value between 0 and 1 (excluded), use a robust quantile
+      of the similarity scores to assign a track to an instance.<br />If equal to 1,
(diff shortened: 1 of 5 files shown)


roomrys (approved):
lgtm

PR #1084, opened 2022-12-15
Flexibly resize input layer of `tf.keras.Model` upon loading trained model

roomrys:
### Description
Flexibly resize the input layer of the `tf.keras.Model` to `(None, None, None, num_channels)` upon loading a trained model. This allows users to use a trained model for inference on any sized video/image.

### Types of changes

- [x] Bugfix
- [ ] New feature
- [ ] Build / CI changes
- [ ] Documentation Update
- [ ] Other (explain)

- #1035

:heart:

### DIFF
--- a/sleap/nn/data/resizing.py
+++ b/sleap/nn/data/resizing.py
@@ -25,8 +25,8 @@ def find_padding_for_stride(
         A tuple of (pad_bottom, pad_right), integers with the number of pixels that the
         image would need to be padded by to meet the divisibility requirement.
     """
-    pad_bottom = (max_stride - (image_height % max_stride)) % max_stride
-    pad_right = (max_stride - (image_width % max_stride)) % max_stride
+    pad_bottom = max_stride - (image_height % max_stride)
+    pad_right = max_stride - (image_width % max_stride)
     return pad_bottom, pad_right
 
 
--- a/sleap/nn/data/resizing.py
+++ b/sleap/nn/data/resizing.py
@@ -25,8 +25,9 @@ def find_padding_for_stride(
         A tuple of (pad_bottom, pad_right), integers with the number of pixels that the
         image would need to be padded by to meet the divisibility requirement.
     """
-    pad_bottom = max_stride - (image_height % max_stride)
-    pad_right = max_stride - (image_width % max_stride)
+    # The outer-most modulo handles edge case when image_height % max_stride == 0
+    pad_bottom = (max_stride - (image_height % max_stride)) % max_stride
+    pad_right = (max_stride - (image_width % max_stride)) % max_stride
     return pad_bottom, pad_right
 
 
--- a/sleap/nn/utils.py
+++ b/sleap/nn/utils.py
@@ -134,7 +134,7 @@ def reset_input_layer(
 ):
     """Returns a copy of `keras_model` with input shape reset to `new_shape`.
 
-    This method was modeified from https://stackoverflow.com/a/58485055.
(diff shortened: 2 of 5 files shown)


sleap/nn/utils.py
```diff
@@ -126,3 +126,52 @@ def match_points(points1: tf.Tensor, points2: tf.Tensor) -> Tuple[tf.Tensor, tf.
...
+        },
+        "inbound_nodes": [],
+    }
+
+    model_config["layers"][1]["inbound_nodes"] = [[[f"{input_layer_name}", 0, 0, {}]]]
+    model_config["input_layers"] = [[f"{input_layer_name}", 0, 0]]
+
+    new_model: tf.keras.Model = keras_model.__class__.from_config(
+        model_config, custom_objects={}
+    )  # Change custom objects if necessary
```
talmo, line 170: I love this function but my only concern is that this might bypass the constructor that is used for custom `tf.keras.Model` subclasses. For example, in our `InferenceModel` subclasses, sometimes we pass more than the underlying `tf.keras.Model` config.

To be safe, I'd just build this with `tf.keras.Model.from_config` instead of `keras_model.__class__.from_config`. This makes it explicit that this function expects a base `tf.keras.Model` typed object, not a subclass for which this initialization method might not be appropriate.

sleap/nn/utils.py
```diff
@@ -126,3 +126,52 @@ def match_points(points1: tf.Tensor, points2: tf.Tensor) -> Tuple[tf.Tensor, tf.
...
+
+    # Iterate over all the layers that we want to get weights from
+    weights = [layer.get_weights() for layer in keras_model.layers[1:]]
+    for layer, weight in zip(new_model.layers[1:], weights):
+        layer.set_weights(weight)
```
talmo, line 175: Is it perhaps less general to assume that the input layer is `keras_model.layers[0]`?

I guess this is baked in above as well, but why do we skip `layers[0]`? In general it should not have weights anyway, but I'm not seeing what's the reason to skip it.

sleap/nn/data/resizing.py
```diff
@@ -25,6 +25,7 @@ def find_padding_for_stride(
         A tuple of (pad_bottom, pad_right), integers with the number of pixels that the
         image would need to be padded by to meet the divisibility requirement.
     """
+    # The outer-most modulo handles edge case when image_height % max_stride == 0
```
talmo, line 28: Thanks for figuring this out!

talmo (approved):
LGTM and I added an integration test, but see comments.

sleap/nn/utils.py
```diff
@@ -148,30 +148,15 @@ def reset_input_layer(
...
-        },
-        "inbound_nodes": [],
-    }
-
-    model_config["layers"][1]["inbound_nodes"] = [[[f"{input_layer_name}", 0, 0, {}]]]
-    model_config["input_layers"] = [[f"{input_layer_name}", 0, 0]]
-
-    new_model: tf.keras.Model = keras_model.__class__.from_config(
+    model_config["layers"][0]["config"]["batch_input_shape"] = new_shape
+    new_model: tf.keras.Model = tf.keras.Model.from_config(
```
roomrys, line 152: Instead of recreating the entire config and specifying the `inbound_nodes` and `input_layers`, I just adjusted the `batch_input_shape`. Both pytest and my own test show that this works as expected (at least on the examples that were tested). @talmo Do you see anything problematic about this?

talmo (approved):
LGTM!!

PR #1142, opened 2023-01-20
Fix body vs symmetry subgraph filtering

talmo:
### Description
Currently, we may not be able to train bottom-up models from the GUI if the skeleton contains symmetry edges that would result in it not being an arborescence.

This is fixed by using only the skeleton subgraph that contains body (non-symmetry) edges to test for whether it is an arborescence.

This PR uses the appropriate subgraph for arborescence testing, as well as fixing how we induced the subgraph views for body vs symmetry edges. Previously, we also had a subtle edge case where calling `Skeleton.graph` would return an empty graph if there were nodes but no body edges. We now return disconnected nodes appropriately.

### Types of changes

- [x] Bugfix
- [ ] New feature
- [ ] Build / CI changes
- [ ] Documentation Update
- [ ] Other (explain)

[list open issues here]

:heart:

### DIFF
--- a/sleap/skeleton.py
+++ b/sleap/skeleton.py
@@ -171,42 +171,39 @@ class Skeleton:
     @property
     def is_arborescence(self) -> bool:
         """Return whether this skeleton graph forms an arborescence."""
-        return nx.algorithms.tree.recognition.is_arborescence(self._graph)
+        return nx.algorithms.tree.recognition.is_arborescence(self.graph)
 
     @property
     def in_degree_over_one(self) -> List[Node]:
-        return [node for node, in_degree in self._graph.in_degree if in_degree > 1]
+        return [node for node, in_degree in self.graph.in_degree if in_degree > 1]
 
     @property
     def root_nodes(self) -> List[Node]:
-        return [node for node, in_degree in self._graph.in_degree if in_degree == 0]
+        return [node for node, in_degree in self.graph.in_degree if in_degree == 0]
 
     @property
     def cycles(self) -> List[List[Node]]:
-        return list(nx.algorithms.simple_cycles(self._graph))
+        return list(nx.algorithms.simple_cycles(self.graph))
 
     @property
     def graph(self):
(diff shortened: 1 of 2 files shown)


sleap/skeleton.py
```diff
@@ -171,42 +171,39 @@ def dict_match(dict1, dict2):
...
-        ]
-        # TODO: properly induce subgraph for MultiDiGraph
-        #   Currently, NetworkX will just return the nodes in the subgraph.
-        #   See: https://stackoverflow.com/questions/16150557/networkxcreating-a-subgraph-induced-from-edges
-        return self._graph.edge_subgraph(edges)
+        """Return a view on the subgraph of body nodes and edges for skeleton."""
+
+        def edge_filter_fn(src, dst, edge_key):
+            edge_data = self._graph.get_edge_data(src, dst, edge_key)
+            return edge_data["type"] == EdgeType.BODY
```
roomrys, line 194: Did you want to make this into a method so we can re-use it?

sleap/skeleton.py
```diff
@@ -171,42 +171,39 @@ def dict_match(dict1, dict2):
...
-        edges = [
-            (src, dst, key)
-            for src, dst, key, edge_type in self._graph.edges(keys=True, data="type")
-            if edge_type == EdgeType.SYMMETRY
-        ]
-        return self._graph.edge_subgraph(edges)
+
+        def edge_filter_fn(src, dst, edge_key):
+            edge_data = self._graph.get_edge_data(src, dst, edge_key)
+            return edge_data["type"] == EdgeType.SYMMETRY
```
roomrys, line 204: Ditto above: Did you want to make this into a method so we can re-use it?

roomrys (approved):
Looks good, small question

talmo:
> Looks good, small question

Honestly it's such a small snippet of mostly networkx code and we exclusively use it for these two cases, so probably not really worth it. We only have two edge types in the skeleton and I can't foresee us having to add more, so I think this is ok since they're both pretty well tested.

PR #1293, opened 2023-05-01
Ensure frames to predict list is unique

roomrys:
### Description
Previously, when predicting on an entire video, we would use the notation `--frames 0,-last_frame_idx`. However, this notation does not properly handle the edge case for single frame videos and results in `--frames 0,0` (which calls inference twice on the same frame). Consequently, our merging code found conflicts in the predictions and existing labels (opting to throw out the existing labels while keeping the predictions). This PR only fixes the double prediction on single frame videos, but the merging code could also be looked at.

### Types of changes

- [x] Bugfix
- [ ] New feature
- [ ] Build / CI changes
- [ ] Documentation Update
- [ ] Other (explain)

[list open issues here]

:heart:

### DIFF
--- a/sleap/gui/learning/runners.py
+++ b/sleap/gui/learning/runners.py
@@ -109,9 +109,9 @@ class VideoItemForInference(ItemForInference):
 
         # -Y represents endpoint of [X, Y) range but inference cli expects
         # [X, Y-1] range (so add 1 since negative).
-        frame_int_list = [i + 1 if i < 0 else i for i in self.frames]
+        frame_int_set = set([i + 1 if i < 0 else i for i in self.frames])
 
-        arg_list.extend(("--frames", ",".join(map(str, frame_int_list))))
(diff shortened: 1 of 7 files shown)


sleap/gui/learning/runners.py
```diff
@@ -109,9 +109,9 @@ def cli_args(self):
 
         # -Y represents endpoint of [X, Y) range but inference cli expects
         # [X, Y-1] range (so add 1 since negative).
-        frame_int_list = [i + 1 if i < 0 else i for i in self.frames]
+        frame_int_set = set([i + 1 if i < 0 else i for i in self.frames])
```
talmo, line 112: Do we maybe need to convert back to list and sort? Sets don't guarantee ordering.

Feel free to ignore if this is not an issue in this case.

talmo (approved):
LGTM conditional on the comment re: set ordering

PR #1130, opened 2023-01-18
Resumable Training

jimzers:
### Description
#1117 PR 1: Add functionality to resume training

### Types of changes

- [ ] Bugfix
- [x] New feature
- [ ] Refactor / Code style update (no logical changes)
- [ ] Build / CI changes
- [ ] Documentation Update
- [ ] Other (explain)

### Does this address any currently open issues?
- #1117 

### Outside contributors checklist

- [ ] Review the [guidelines for contributing](https://github.com/talmolab/sleap/blob/develop/docs/CONTRIBUTING.md) to this repository
- [ ] Read and sign the [CLA](https://github.com/talmolab/sleap/blob/develop/sleap-cla.pdf) and add yourself to the [authors list](https://github.com/talmolab/sleap/blob/develop/AUTHORS)
- [ ] Make sure you are making a pull request against the **develop** branch (not *main*). Also you should start *your branch* off *develop*
- [ ] Add tests that prove your fix is effective or that your feature works
- [ ] Add necessary documentation (if appropriate)

#### Thank you for contributing to SLEAP!
:heart:

PR #1314, opened 2023-05-18
Fix GUI resume training

roomrys:
### Description
After attempting a few rounds of training via the GUI, we were a bit surprised to find that the results when resuming training did not perform better than the results for training from scratch. When digging into this, we noticed that the `base_checkpoint` parameter in the training_config.json file was always null even when directly specifying a file path to the `base_checkpoint`. It turns out that the CLI arg `--base_checkpoint` was always overwriting the parameter set in the training config (making the GUI resume training useless). 

This PR ensure that the `base_checkpoint` is only overwritten if a non-default value is specified for `--base_checkpoint`.

Related: #1130

### Types of changes

- [x] Bugfix
- [ ] New feature
- [ ] Build / CI changes
- [ ] Documentation Update
- [ ] Other (explain)

[list open issues here]

:heart:

### DIFF
--- a/sleap/nn/training.py
+++ b/sleap/nn/training.py
@@ -1844,9 +1844,10 @@ def create_trainer_using_cli(args: Optional[List] = None):
     parser.add_argument(
         "--base_checkpoint",
         type=str,
+        default=None,
         help=(
             "Path to base checkpoint (directory containing best_model.h5) to resume "
-            "training from."
+            "training from. Default is None."
         ),
     )
     parser.add_argument(
@@ -1940,7 +1941,8 @@ def create_trainer_using_cli(args: Optional[List] = None):
     if len(args.video_paths) == 0:
         args.video_paths = None
 
-    job_config.model.base_checkpoint = args.base_checkpoint
+    if args.base_checkpoint is not None:
+        job_config.model.base_checkpoint = args.base_checkpoint
 
     logger.info("Versions:")
     sleap.versions()
--- a/tests/nn/test_training.py
+++ b/tests/nn/test_training.py
@@ -316,43 +316,47 @@ def test_train_cropping(
     )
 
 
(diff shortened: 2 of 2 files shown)


talmo:
I think we need to add one more check here on this fix to address item 4 here: https://github.com/talmolab/sleap/discussions/1291#discussioncomment-5951648

>> 4. Iin the centered_instance.json, I notice that the `"training_inds"` list shows only the prior trained model's list of frame numbers and **does not** include any newly annotated frame numbers. _# Is this what I should be seeing in the json file when retraining with new training data?_
> 
> Thanks for pointing that out! I'm not sure we had noticed that previously. The new `centered_instance.json` probably shouldn't have the list of frame numbers at all and that may indeed be leading to the trainer not using the new labels! We'll look into it as part of the bug fix investigation mentioned above.

Namely, I'm concerned that this part of the `DataReaders` is kicking in:
https://github.com/talmolab/sleap/blob/d05e2bf524cdb9813926a841886725644f00a218/sleap/nn/training.py#L191-L199
which results in any new labeled frames to be not used in training when resuming from a previous checkpoint.

This would also lead to issues when resuming from a checkpoint used in another labels project entirely which may have different numbers of labeled frames.

I know the old config isn't used in the `Trainer` (only the `best_model.h5` during model initialization), but it _will_ carry over when doing this from the GUI training config editor.

talmo (approved):

roomrys:
The above would only apply if `split_by_inds == True`. 
https://github.com/talmolab/sleap/blob/d05e2bf524cdb9813926a841886725644f00a218/sleap/nn/training.py#L173-L178
Thus, I created another PR (#1324) which allows users the option to "Split by Indices" if they would like to rerun the exact same experiment in the GUI (to measure any noise, or exited early previously). However, this checkbox is likely most useful for de-selecting "Split by Indices" if selected in the original config (while being transparent).

Issue #1375, opened 2023-07-06
Remove Videos in Batch


talmo:
### Discussed in https://github.com/talmolab/sleap/discussions/1363

<div type='discussions-op-text'>

<sup>Originally posted by **daskandalis** June 23, 2023</sup>
Currently videos have to be removed one at a time. Videos can be added in a batch, it would be helpful to have a function to remove them in a batch as well.</div>

roomrys:
This should be addressed in [SLEAP 1.3.2](https://github.com/talmolab/sleap/releases/tag/v1.3.2).

PR #1382, opened 2023-07-10
Add option to remove videos in batch

gitttt-1234:
### Description
Added an option to remove videos in batch.
Added Test cases to test different scenarios for removing multiple videos at the same time

### Types of changes

- [ ] Bugfix
- [.] New feature
- [ ] Build / CI changes
- [ ] Documentation Update
- [ ] Other (explain)

Remove Videos in Batch #1375

- [x] Add necessary documentation (if appropriate)

:heart:

### DIFF
--- a/sleap/gui/commands.py
+++ b/sleap/gui/commands.py
@@ -1843,27 +1843,29 @@ class RemoveVideo(EditCommand):
     def do_action(context: CommandContext, params: dict):
         videos = context.labels.videos.copy()
         row_idxs = context.state["selected_batch_video"]
-        
+
         # Remove selected videos in the project
         for idx in row_idxs:
 
-            #check if video to be deleted is the current state video
-            if videos[idx]==context.state["video"]:
+            # check if video to be deleted is the current state video
+            if videos[idx] == context.state["video"]:
                 if len(context.labels.videos):
-                    context.state["video"] = context.labels.videos[list(set(range(len(videos)))-set(row_idxs))[-1]]
+                    context.state["video"] = context.labels.videos[
+                        list(set(range(len(videos))) - set(row_idxs))[-1]
+                    ]
                 else:
                     context.state["video"] = None
-                
+
             context.labels.remove_video(videos[idx])
 
     @staticmethod
     def ask(context: CommandContext, params: dict) -> bool:
         videos = context.labels.videos.copy()
         row_idxs = context.state["selected_batch_video"]
-        video_file_names=[]
+        video_file_names = []
         total_num_labeled_frames = 0
         for idx in row_idxs:
-        
+
             video = videos[idx]
             if video is None:
                 return False
@@ -1872,11 +1874,13 @@ class RemoveVideo(EditCommand):
             n = len(context.labels.find(video))
 
             if n > 0:
-                total_num_labeled_frames+=n
-                video_file_names.append(f"{video}".split(", shape")[0].split("filename=")[-1].split("/")[-1])
+                total_num_labeled_frames += n
+                video_file_names.append(
+                    f"{video}".split(", shape")[0].split("filename=")[-1].split("/")[-1]
+                )
 
         # Warn if there are labels that will be deleted
-        if len(video_file_names)>=1:
+        if len(video_file_names) >= 1:
             response = QtWidgets.QMessageBox.critical(
                 context.app,
                 "Removing video with labels",
--- a/sleap/gui/commands.py
+++ b/sleap/gui/commands.py
@@ -428,10 +428,6 @@ class CommandContext:
         """Removes selected video from project."""
         self.execute(RemoveVideo)
 
-    def removeAllVideos(self):
-        '''Removes All Videos from project'''
-        self.execute(RemoveAllVideos)
-
     def openSkeletonTemplate(self):
         """Shows gui for loading saved skeleton into project."""
         self.execute(OpenSkeleton, template=True)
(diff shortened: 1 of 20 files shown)


roomrys (changes requested):
Would it be possible to add a test in the test_commands.py file that tests the `do_action` method of `RemoveVideos`? It is fine to spoof the row indices for the test.

sleap/gui/commands.py
```diff
@@ -1841,40 +1841,46 @@ class RemoveVideo(EditCommand):
...
+        videos = context.labels.videos.copy()
+        row_idxs = context.state["selected_batch_video"]
+        # Remove selected videos in the project
+        for idx in row_idxs:
+            context.labels.remove_video(videos[idx])
+
+        if len(context.labels.videos):
+            context.state["video"] = context.labels.videos[-1]
+        else:
+            context.state["video"] = None
```
roomrys, line 1853: Check that the current video was removed before changing the state

sleap/gui/dataviews.py
```diff
@@ -370,6 +375,11 @@ def getSelectedRowItem(self) -> Any:
         not the converted dict.
         """
         idx = self.currentIndex()
+
+        if self.multiple_selection:
+            idx_temp = set([x.row() for x in self.selectedIndexes()])
+            self.state[f"selected_batch_"+self.row_name] = idx_temp
```
roomrys, line 381:
```suggestion
            self.state[f"selected_batch_{self.row_name}"] = idx_temp
```

tests/gui/test_commands.py
```diff
@@ -89,6 +89,28 @@ def test_get_new_version_filename():
...
+        centered_pair_vid: Video,
+        small_robot_3_frame_vid: Video,
+):
+    labels = centered_pair_predictions.copy()
+    labels.add_video(small_robot_mp4_vid)
+    labels.add_video(centered_pair_vid)
+    labels.add_video(small_robot_3_frame_vid)
+
+    all_videos = list(labels.videos)
+    assert len(all_videos) == 3
```
roomrys, line 104: Fix test

sleap/gui/commands.py
```diff
@@ -1841,40 +1841,46 @@ class RemoveVideo(EditCommand):
...
 
         # Warn if there are labels that will be deleted
-        if n > 0:
+        if len(video_file_names)>=1:
             response = QtWidgets.QMessageBox.critical(
                 context.app,
                 "Removing video with labels",
-                f"{n} labeled frames in this video will be deleted, "
-                "are you sure you want to remove this video?",
+                f"Labeled frames in {', '.join(video_file_names)} will be deleted, "
```
roomrys, line 1877: I liked that you had the number of labeled frames to be deleted before

roomrys (changes requested):

sleap/gui/commands.py
```diff
@@ -1843,20 +1843,23 @@ class RemoveVideo(EditCommand):
...
-                if len(context.labels.videos):
+        # check if video to be deleted is the current state video
+        if context.state["video"] in videos_to_be_removed:
+            if len(context.labels.videos):
+                try:
                     context.state["video"] = context.labels.videos[
                         list(set(range(len(videos))) - set(row_idxs))[-1]
                     ]
-                else:
+                except:
```
talmo, line 1859: Using a plain `try-except` without specifying the particular exception you're expecting to handle is a [common anti-pattern](https://realpython.com/the-most-diabolical-python-antipattern/).

Doing so is quick to write and handle complex control flow scenarios, but it has the side effect of obscuring errors that we didn't mean to catch. Instead I think what you want to do here is use a more specific error type, like:

```py
try:
    ...
except IndexError:
    # Comment on what the specific case you're catching is and what we're doing to handle it.
    ...
```

[Here's a list of common exceptions](https://docs.python.org/3/library/exceptions.html).

In this particular case, you could just do a test for whether the new video is available and just wrap this in an `if-else` block.

It also helps to document under what circumstances the error you're catching is raised, and what the new behavior is if that happens.

Like right now it looks like it defaults to switching to the last video, but I don't see that described anywhere.

tests/gui/test_commands.py
talmo: Don't forget to resolve this conversation from the previous review if already addressed.

tests/gui/widgets/test_docks.py
```diff
@@ -11,22 +11,25 @@
...
-    labels.add_video(centered_pair_vid)
     labels.add_video(small_robot_3_frame_vid)
+    labels.add_video(centered_pair_vid)
+    labels.add_video(small_robot_mp4_vid)
     assert len(labels.videos) == 4
 
     # Create the dock
     main_window = MainWindow()
-    main_window.labels = labels
+    main_window.commands.loadLabelsObject(labels)
```
talmo, line 32: Since we spent some time troubleshooting this, it's worth adding a comment as to why we're using it via the command as opposed to just setting the attribute for future reference in case we need to write these types of tests again.

talmo (changes requested):

sleap/gui/commands.py
gitttt-1234: sure, I'll wrap it in an if-else condition

tests/gui/widgets/test_docks.py
gitttt-1234: sure, I'll add a comment

sleap/gui/dataviews.py
```diff
@@ -301,6 +301,7 @@ def __init__(
         is_sortable: bool = False,
         is_activatable: bool = False,
         ellipsis_left: bool = False,
+        multiple_selection=False,
```
talmo, line 304: Add type hint for consistency:

```suggestion
        multiple_selection: bool = False,
```

tests/gui/test_commands.py
```diff
@@ -90,6 +91,38 @@ def test_get_new_version_filename():
...
+):
+    def ask(obj: RemoveVideo, context: CommandContext, params: dict) -> bool:
+        return True
+
+    RemoveVideo.ask = ask
+
+    labels = centered_pair_predictions.copy()
+    labels.add_video(small_robot_mp4_vid)
+    labels.add_video(centered_pair_vid)
+    # labels.add_video(small_robot_3_frame_vid)
```
talmo, line 108: remove unused comment

talmo (approved):
LGTM! Make a couple of minor fixes, add a description to the PR, and go ahead and squash merge when all the checks pass.

PR #1435, opened 2023-08-02
Redraw overlays on plot change

roomrys:
### Description
Previously, overlays had an open reference even after they were cleared from the player. This PR Ensures that when `QtVideoPlayer.reset` is called, the `changedPlot` signal is emitted which handles overlays and gui states.
> Note: When a video is removed via the `RemoveVideo` command, this will call the `MainWindow.on_data_update` function with a topic of `UpdateTopic.frame`. The `MainWindow.on_data_update` will then be called a second time with the `UpdateTopic.frame` if the video removed is the last video (which then calls `QtVideoPlayer.reset`). Although redundant, this will not be an expensive operation since most the data from the slp will be removed at this point (all `LabeledFrames`, `Suggestions`, and `Videos`).

### Types of changes

- [x] Bugfix
- [ ] New feature
- [ ] Build / CI changes
- [ ] Documentation Update
- [ ] Other (explain)

[list open issues here]

:heart:

### DIFF
--- a/sleap/gui/app.py
+++ b/sleap/gui/app.py
@@ -1222,6 +1222,8 @@ class MainWindow(QMainWindow):
         # Select instance if there was already selection
         if selected_inst is not None:
             player.view.selectInstance(selected_inst)
+        else:
+            self.state["instance"] = None
 
         if self.state["fit"]:
             player.zoomToFit()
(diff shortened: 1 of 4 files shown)


shrivaths16 (approved):
Looks good.

PR #1445, opened 2023-08-09
Add model folder to the unzip path

roomrys:
### Description
Not sure how this crept in unnoticed for so long... but when unzipping a model path, the directory that contains all the model info was not being appended to the path. This PR appends the directory.

### Types of changes

- [x] Bugfix
- [ ] New feature
- [ ] Build / CI changes
- [ ] Documentation Update
- [ ] Other (explain)

- #1443

:heart:

### DIFF
--- a/sleap/nn/inference.py
+++ b/sleap/nn/inference.py
@@ -4824,6 +4824,7 @@ def load_model(
         # Uncompress ZIP packaged models.
         tmp_dirs = []
         for i, model_path in enumerate(model_paths):
+            mp = Path(model_path)
             if model_path.endswith(".zip"):
                 # Create temp dir on demand.
                 tmp_dir = tempfile.TemporaryDirectory()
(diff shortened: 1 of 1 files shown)


talmo:
Actually we should probably support both, checking for the existence of `best_model.h5`.

We have models that have basically this organization:

```
model_name.zip
model_name.zip/best_model.h5
model_name.zip/training_config.json
...
```
which is what the rarely used [`Trainer.package()`](https://github.com/talmolab/sleap/blob/d61a1848742bae7b239a5337eceb50c59e6749ab/sleap/nn/training.py#L1000-L1007) produces.

But if zipping manually, it's way easier to just zip the parent folder than to zip its **contents** and pull that zip file out. In this case, the organization is more like:

```
model_name.zip
model_name.zip/model_name/best_model.h5
model_name.zip/model_name/training_config.json
...
```

We have some example notebooks that use the former, so if we're adding support for the latter, we should make sure we can do both.

davidasamy (approved):
lgtm

Issue #1308, opened 2023-05-10
Best effort tracking with hard cap on max number of tracks


talmo:
Placeholder issue for suggestions in #1288 and #797.

We should add a max tracks constraint to the tracking candidate buffer and stickiness to enforce a global max number of tracks, like this:  https://gist.github.com/talmo/a5415218c194536092cda10f897f8cad

talmo:
Some more fleshing out of this feature:

The key logic we need to modify is here:

https://github.com/talmolab/sleap/blob/e0eebb260c9475571d14eb4d0ce570de402f12b3/sleap/nn/tracking.py#L505-L510

This part of the `Tracker.track()` method generates candidates (instances that already have track assignments) for association with new incoming instances that do not have tracks.

We only have two candidate makers: `SimpleCandidateMaker` and `FlowCandidateMaker`. The former just returns the instances in the window of frames in the current track matching queue:

https://github.com/talmolab/sleap/blob/e0eebb260c9475571d14eb4d0ce570de402f12b3/sleap/nn/tracking.py#L315-L334

The `FlowCandidateMaker` does something similar but also applies optical flow to update the locations of those candidate instances.

Either way, what we need to do is to create new candidate makers that operate on multiple queues (1 per track, up to max tracks) rather than a single queue which has all tracks combined.

PR #1447, opened 2023-08-10
Limit max tracks via track-local queues

shrivaths16:
### Description
While performing inference based tracking, the number of tracks that are generated needs to be capped in order to avoid creating multiple irrelevant/redundant tracks. 

We implement a new mode of tracking here to address this by enabling track-local queues to store candidates for matching. This allows for setting the global maximum number of tracks, irrespective of how large the gaps between detections of a track are.

This also exposes that functionality to the `load_model` API, the CLI, and the GUI, where now superseded options such as pre-tracking culling have been removed.

### Types of changes

- [ ] Bugfix
- [x] New feature
- [ ] Build / CI changes
- [ ] Documentation Update
- [ ] Other (explain)

- #1308

:heart:

### Summary by CodeRabbit

- New Feature: Introduced maximum tracking and maximum number of tracks options to the `Tracker` class, enhancing its flexibility and functionality.
- New Feature: Added support for these new tracking options in the `load_model` and `_make_tracker_from_cli` functions.
- Test: Expanded test coverage with new test cases for maximum tracking with large gaps, small gaps on both tracks, and extra detections.
- Documentation: Updated CLI documentation to include new tracking options such as `--tracking.max_tracking`, `--tracking.max_tracks`, and `--tracking.robust`.
- Refactor: Adjusted tracker naming conventions based on the presence of `max_tracking` and `max_tracks` parameters.

### DIFF
--- a/sleap/instance.py
+++ b/sleap/instance.py
@@ -333,9 +333,6 @@ class Track:
         """
         return attr.asdict(self) == attr.asdict(other)
 
-    def __hash__(self) -> int:
-        return hash(self.spawned_on, self.name)
-
 
 # NOTE:
 # Instance cannot be a slotted class at the moment. This is because it creates
--- a/sleap/instance.py
+++ b/sleap/instance.py
@@ -333,6 +333,9 @@ class Track:
         """
         return attr.asdict(self) == attr.asdict(other)
 
+    def __hash__(self) -> int:
+        return hash((self.spawned_on, self.name))
+
 
 # NOTE:
 # Instance cannot be a slotted class at the moment. This is because it creates
--- a/sleap/instance.py
+++ b/sleap/instance.py
@@ -333,9 +333,6 @@ class Track:
         """
         return attr.asdict(self) == attr.asdict(other)
 
-    def __hash__(self) -> int:
-        return hash((self.spawned_on, self.name))
-
 
 # NOTE:
 # Instance cannot be a slotted class at the moment. This is because it creates
--- a/sleap/instance.py
+++ b/sleap/instance.py
@@ -333,6 +333,9 @@ class Track:
         """
         return attr.asdict(self) == attr.asdict(other)
 
+    def __hash__(self) -> int:
+        return hash(self.spawned_on, self.name)
+
 
 # NOTE:
 # Instance cannot be a slotted class at the moment. This is because it creates
--- a/sleap/nn/tracking.py
+++ b/sleap/nn/tracking.py
@@ -345,6 +345,7 @@ class SimpleCandidateMaker:
 class SimpleMaxTracksCandidateMaker:
     """Class to generate instances based on the maximum number of tracks from prior frames."""
 
+    max_tracks: int = None
     min_points: int = 0
 
     @property
@@ -361,7 +362,7 @@ class SimpleMaxTracksCandidateMaker:
         candidate_instances = []
         number_of_tracks = 0
         for track in track_matching_queue_dict.keys():
-            if track and number_of_tracks <= max_tracks:
+            if track and number_of_tracks <= self.max_tracks:
                 number_of_tracks += 1
                 for matched_item in track:
                     ref_t, ref_instances = matched_item.t, matched_item.instances_t
@@ -447,7 +448,7 @@ class Tracker(BaseTracker):
             0.95 is a good value.
     """
 
-    max_tracks: int
+    max_tracks: int = None
     track_window: int = 5
     similarity_function: Optional[Callable] = instance_similarity
     matching_function: Callable = greedy_matching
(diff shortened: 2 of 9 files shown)


sleap/nn/tracking.py
```diff
@@ -311,6 +318,119 @@ def flow_shift_instances(
...
+        candidate_instances = []
+
+        # Prune old shifted instances to save time and memory
+        self.prune_shifted_instances(t)
+        number_of_tracks = 0
+        # -> List[ShiftedInstance]
+        def get_ref_instances(r_t, r_img) -> List[InstanceType]:
+            instances = []
+            for track, matched_items in track_matching_queue_dict.items():
+                instances + [
```
roomrys, line 375:
```suggestion
                instances += [
```
Add test where this fails before adding fix

sleap/nn/tracking.py
```diff
@@ -334,9 +454,40 @@ def get_candidates(
...
+        self,
+        track_matching_queue_dict: dict(),
+        *args,
+        **kwargs,
+    ) -> List[InstanceType]:
+        # Create set of matchable candidate instances from each track for max number of tracks.
+        candidate_instances = []
+        number_of_tracks = 0
+        for track, matched_instances in track_matching_queue_dict.items():
+            if track and number_of_tracks <= self.max_tracks:
```
roomrys, line 478: We shouldn't need to check for the `track` at this point since a `track is None` should not get added to the `track_matching_queue_dict`

sleap/nn/tracking.py
```diff
@@ -443,7 +600,11 @@ def _init_matching_queue(self):
         return deque(maxlen=self.track_window)
 
     def reset_candidates(self):
-        self.track_matching_queue = deque(maxlen=self.track_window)
+        if self.max_tracking and self.track_matching_queue_dict:
+            for track, candidates in self.track_matching_queue_dict.items():
+                candidates = deque(maxlen=self.track_window)
+        else:
+            self.track_matching_queue = deque(maxlen=self.track_window)
```
roomrys, line 607: Set the item in `track_matching_queue` itself, this might not actually update the `track_matching_queue`
```python
In [1]: d = {1: 1}

In [2]: for k, v in d.items():
   ...:     v = 3
   ...:

In [3]: d
Out[3]: {1: 1}
```

roomrys (dismissed):
Small review from subset group call

sleap/instance.py
```diff
@@ -333,6 +333,9 @@ def matches(self, other: "Track"):
         """
         return attr.asdict(self) == attr.asdict(other)
 
+    def __hash__(self) -> int:
+        return hash((self.spawned_on, self.name))
+
```
talmo, line 338: Danger zone: let's not override whatever attrs is doing with hashing. [See these docs for reference](https://www.attrs.org/en/stable/hashing.html).

We currently have this class set to `eq=False`, which means it's hashing by object identity. The equivalent of what's being done here would be to set `frozen=True`, which will hash by value. But as you can see above, we have a `matches` method that does basically that, which suggests to me that some other part of the code probably expects it to be hashable by identity instead.

sleap/nn/tracking.py
```diff
@@ -132,6 +139,51 @@ class FlowCandidateMaker:
     def uses_image(self):
         return True
 
+    def get_shifted_instances_from_earlier_time(
+        self, ref_t: int, ref_img: np.ndarray, ref_instances: List[InstanceType], t: int
+    ) -> (np.ndarray, List[InstanceType]):
+        """Check if shifted instance was computed at earlier time and return instances and corresponding image."""
```
talmo, line 145: - 88 characters max line length
- Add args to docstring

sleap/nn/tracking.py
```diff
@@ -132,6 +139,51 @@ class FlowCandidateMaker:
...
+
+    def get_shifted_instances(
+        self,
+        ref_instances: List[InstanceType],
+        ref_img: np.ndarray,
+        ref_t: int,
+        img: np.ndarray,
+        t: int,
+    ) -> List[ShiftedInstance]:
+        """Returns a list of shifted instances and saves the shifted instances if necessary."""
```
talmo, line 164: Args in docstring

sleap/nn/tracking.py
```diff
@@ -311,6 +339,79 @@ def flow_shift_instances(
         return shifted_instances
 
 
+@attr.s(auto_attribs=True)
+class FlowMaxTracksCandidateMaker(FlowCandidateMaker):
+    """Class for producing optical flow shift matching candidates with cap on the number of tracks.
+
+    Attributes:
+        max_tracks: The maximum number of tracks that needs to be maintained in order to avoid redundant/irrelevant tracks.
```
talmo, line 347: 88 characters max line width

sleap/nn/tracking.py
```diff
@@ -443,7 +576,11 @@ def _init_matching_queue(self):
         return deque(maxlen=self.track_window)
 
     def reset_candidates(self):
-        self.track_matching_queue = deque(maxlen=self.track_window)
+        if self.max_tracking and self.track_matching_queue_dict:
```
talmo, line 579: Do we need to check for `self.track_matching_queue_dict`? This will be initialized by the factory method in the attribute definition anyway.

sleap/nn/tracking.py
```diff
@@ -622,6 +804,9 @@ def make_tracker_by_name(
         # Kalman filter options
         kf_init_frame_count: int = 0,
         kf_node_indices: Optional[list] = None,
+        # Max tracking options
+        max_tracks: int = None,
```
talmo, line 808: `Optional[int] = None`?

talmo (changes requested):
Minor changes

talmo (approved):

Issue #1476, opened 2023-09-05
Failure to Start GUI


Nike-Nay:
# Bug description

I have installed sleap using mambaforge and installed it via the 'conda package' instructions at https://sleap.ai/installation.html
I believe everything is installed correctly as I can activate sleap in the terminal.
However, when I attempt to load up the GUI using 'sleap-label' it returns the following:
FileNotFoundError: [Errno 2] No such file or directory: 'c:\\users\\asus\\mambaforge\\envs\\sleap\\lib\\site-packages\\tensorflow_estimator-2.6.0.dist-info\\METADATA'

I am not sure why there is no METADATA file. Is this an issue with the downloaded package or an issue on my system?

## Your personal set up

- OS: Microsoft Windows 10 Home

- Version(s): SLEAP v1.3.1, python 3.7.12

- SLEAP installation method (listed [here](https://sleap.ai/installation.html#)):
  - [x] [Conda from package](https://sleap.ai/installation.html#conda-package)
  - [ ] [Conda from source](https://sleap.ai/installation.html#conda-from-source)
  - [ ] [pip package](https://sleap.ai/installation.html#pip-package)
  - [ ] [Apple Silicon Macs](https://sleap.ai/installation.html#apple-silicon-macs)

<details><summary>Environment packages</summary>

```
# paste output of `pip freeze` or `conda list` here
(sleap) C:\Users\asus>sleap-label
Traceback (most recent call last):
  File "C:\Users\asus\mambaforge\envs\sleap\lib\site-packages\pkg_resources\__init__.py", line 573, in _build_master
    ws.require(__requires__)
  File "C:\Users\asus\mambaforge\envs\sleap\lib\site-packages\pkg_resources\__init__.py", line 891, in require
    needed = self.resolve(parse_requirements(requirements))
  File "C:\Users\asus\mambaforge\envs\sleap\lib\site-packages\pkg_resources\__init__.py", line 782, in resolve
    raise VersionConflict(dist, req).with_context(dependent_req)
pkg_resources.ContextualVersionConflict: (importlib-metadata 4.13.0 (c:\users\asus\mambaforge\envs\sleap\lib\site-packages), Requirement.parse('importlib-metadata<4.3; python_version < "3.8"'), {'hdmf'})

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\Users\asus\mambaforge\envs\sleap\lib\site-packages\pkg_resources\__init__.py", line 3031, in _dep_map
[...]

Issue #1390, opened 2023-07-17
Add instructions for CUDA issues with new TensorFlow versions


talmo:
We should add something in the docs about adding the CUDA paths manually to get around the annoying [TensorFlow-not-finding-CUDA issue](https://github.com/tensorflow/tensorflow/issues/52988):

```
conda activate sleap
conda env config vars set LD_PRELOAD=$CONDA_PREFIX/lib/libcudart.so:$CONDA_PREFIX/lib/libcublas.so:$CONDA_PREFIX/lib/libcublasLt.so:$CONDA_PREFIX/lib/libcufft.so:$CONDA_PREFIX/lib/libcurand.so:$CONDA_PREFIX/lib/libcusolver.so:$CONDA_PREFIX/lib/libcusparse.so:$CONDA_PREFIX/lib/libcudnn.so
conda deactivate
conda activate sleap

# ... run things as usual
```

-----
**Original discussion:**

Hi @cxrodgers,

> I'm unable to update nvidia drivers because I don't have admin privileges and in any case this is a shared cluster so they won't want to break other people's stuff.

I figured... I imagine they'll want to bring that up to >525 at some point soon since it's required for newer CUDA versions. Probably not the culprit here though.

> Previous clusters I've used had some kind of "head node" where you could submit jobs but couldn't see any GPUs. This one is set up differently, I can see all 4 GPUs as soon as I log in. But they told me never to use a GPU outside of a slurm job or I'd get kicked off the cluster. So when I submit a slurm job, then `sleap.system_setup` can only see the GPU that I reserve, and that's when the problem occurs.

Yeah weird that they have GPUs on the head node at all! Maybe they're just reusing a compute node?

In any case, assuming the compute nodes where the SLURM jobs actually get run is also using A6000s, I don't think this would be the issue.

This brings us back to the CUDA/cuDNN version. Because we install it via conda, the appropriate versions should be used, but if you're on TensorFlow 2.7.0, you might be running into [this issue where newer versions of TensorFlow fail to detect the conda cuda/cudnn libraries](https://github.com/tensorflow/tensorflow/issues/52988). Normally you'd get some errors about failing to load cuda/cudnn binaries, but because you have a system-level cuda/cudnn installed, it might be falling back to that and generating the weird driver version errors we're seeing.

Try adding these lines in your SLURM script as a workaround:
```
conda activate sleap
[...]

PR #1496, opened 2023-09-12
Set `LD_LIBRARY_PATH` on `mamba activate`

roomrys:
### Description
Some Linux users have been unable to utilize the GPU. Modifying some installation instructions from `tensorflow`, this PR sets the `LD_LIBRARY+PATH` on `mamba activate` (if installing SLEAP via the conda package) and gives instructions for setting the `LD_LIBRARY_PATH` manually (once and then automated on `mamba activate`) for conda from source.

Related:
- #1390

### Types of changes

- [ ] Bugfix
- [x] New feature
- [ ] Build / CI changes
- [x] Documentation Update
- [ ] Other (explain)

- #1476

:heart:

### Summary by CodeRabbit

- New Feature: Switched from `conda` to `mamba` for package management, improving the speed and reliability of environment creation and package installation.
- Bug Fix: Added a shell script that sets the `LD_LIBRARY_PATH` for Linux users with NVIDIA GPUs, resolving issues with GPU utilization.
- Documentation: Updated the installation instructions in `installation.md` to reflect the switch to `mamba` and added guidance for Linux users with NVIDIA GPUs on how to set up their environment correctly.

### DIFF
--- a/.conda/build.sh
+++ b/.conda/build.sh
@@ -13,3 +13,11 @@ pip install --no-cache-dir -r ./requirements.txt
 # Install sleap itself. This does not install the requirements, but will list which 
 # requirements are missing (see "install_requires") when user attempts to install.
 python setup.py install --single-version-externally-managed --record=record.txt
+
+# Copy the activate scripts to $PREFIX/etc/conda/activate.d.
+# This will allow them to be run on environment activation.
+export CHANGE=activate
+
+mkdir -p "${PREFIX}/etc/conda/${CHANGE}.d"
+ls "${RECIPE_DIR}"
+cp "${RECIPE_DIR}/${PKG_NAME}_${CHANGE}.sh" "${PREFIX}/etc/conda/${CHANGE}.d/${PKG_NAME}_${CHANGE}.sh"
--- a/.conda/sleap_activate.sh
+++ b/.conda/sleap_activate.sh
@@ -1,0 +1,4 @@
+#!/bin/sh
+
+# Help CUDA find GPUs!
+export LD_LIBRARY_PATH=$CONDA_PREFIX/lib:$LD_LIBRARY_PATH
--- a/environment.yml
+++ b/environment.yml
@@ -46,16 +46,4 @@ dependencies:
 
     - pip:  
         - "--editable=.[conda_dev]"
-
-# If you are on Linux, have a NVIDIA GPU, and are getting the warning:
-
-# W tensorflow/stream_executor/platform/default/dso_loader.cc:64] Could not load dynamic 
-# library 'libcudart.so.11.0'; dlerror: libcudart.so.11.0: cannot open shared object 
-# file: No such file or directory
-
-# then activate the environment and run the commands:
-
-# mkdir -p $CONDA_PREFIX/etc/conda/activate.d
-# echo '#!/bin/sh' >> $CONDA_PREFIX/etc/conda/activate.d/libcudart_activate.sh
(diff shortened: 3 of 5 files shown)


.conda/build.sh
```diff
@@ -12,4 +12,12 @@ pip install --no-cache-dir -r ./requirements.txt
...
\ No newline at end of file
+python setup.py install --single-version-externally-managed --record=record.txt
+
+# Copy the activate scripts to $PREFIX/etc/conda/activate.d.
+# This will allow them to be run on environment activation.
+export CHANGE=activate
+
+mkdir -p "${PREFIX}/etc/conda/${CHANGE}.d"
+ls "${RECIPE_DIR}"
+cp "${RECIPE_DIR}/${PKG_NAME}_${CHANGE}.sh" "${PREFIX}/etc/conda/${CHANGE}.d/${PKG_NAME}_${CHANGE}.sh"
```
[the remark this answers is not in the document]
roomrys: The build just fails and gives us an error message if these lines end in failure.

.conda/sleap_activate.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+
+# Help CUDA find GPUs!
+export LD_LIBRARY_PATH=$CONDA_PREFIX/lib:$LD_LIBRARY_PATH
```
talmo, line 4: Are you sure this will work? I thought the cuda binaries were stored in the `$CONDA_PREFIX/envs/{env_name}/lib` or something.

Like, one blocker for upgrading to newer tensorflow was that we couldn't dynamically set this since users would have different env names.

If it's really in `$CONDA_PREFIX/lib` and it's not just hardcoding the path from the computer you built in (would Jinja do this?), then that's awesome and would allow us to move up to tf 2.9 I think!

If it _doesn't_ work, then it might be ok to just include the extra line that people have to run in the install docs?

environment.yml
```diff
@@ -46,3 +46,16 @@ dependencies:
...
+# W tensorflow/stream_executor/platform/default/dso_loader.cc:64] Could not load dynamic 
+# library 'libcudart.so.11.0'; dlerror: libcudart.so.11.0: cannot open shared object 
+# file: No such file or directory
+
+# then activate the environment and run the commands:
+
+# mkdir -p $CONDA_PREFIX/etc/conda/activate.d
+# echo '#!/bin/sh' >> $CONDA_PREFIX/etc/conda/activate.d/libcudart_activate.sh
+# echo 'export LD_LIBRARY_PATH=$CONDA_PREFIX/lib:$LD_LIBRARY_PATH' >> $CONDA_PREFIX/etc/conda/activate.d/libcudart_activate.sh
+# source $CONDA_PREFIX/etc/conda/activate.d/libcudart_activate.sh
```
talmo, line 61: Possibly move to docs?

roomrys: This is also in the docs under GPU support. If you dislike it here I can remove?

.conda/sleap_activate.sh
roomrys:
```bash
(nb4) talmolab@talmolab-01-ubuntu:~/sleap-estimates-animal-poses/pull-requests/sleap$ echo $CONDA_PREFIX
/home/talmolab/micromamba/envs/nb4
```

!!!

roomrys: >If it's really in $CONDA_PREFIX/lib and it's not just hardcoding the path from the computer you built in (would Jinja do this?), then that's awesome and would allow us to move up to tf 2.9 I think!

I think that might be the point of using `cp` now that you mention it (instead of `echo ... >> ...`).

environment.yml
roomrys: Removed

Issue #1065, opened 2022-12-01
Prefer user instances when "Copy Prior Frame"


Deruijter:
Hi,

I'm fairly new to SLEAP so I'm not sure I'm doing something incorrectly or that this is unintended behavior.

I've manually labeled about 240 frames (~480 animals) in my videos. I'm now testing to run inference on unlabeled video frames. I'm starting with predicting an individual frame that takes place after a frame that was: first predicted, and then manually corrected. 

I was not satisfied with the inferred frame as the positioning was more off than I expected (particularly because the animal hardly moved). I then proceeded using the "copy prior frame" option to label the frame myself. To my surprise the new instance was different than that of the prior frame. It actually has the same shape as the original prediction of the prior frame, but not the corrected instance. I'm guessing the same is happening while trying to infer this frame.

Is it correct that "post prediction -corrected- instances" are somehow ignored in this regard? I was hoping to kick start the inference in the correct direction by starting with a few manually corrected frames.

Note that predictions + corrections of these frames were done in version 1.2.5. I now updated to 1.2.9 but the "copy prior frame" problem seems to persist.

Some images to show what is going on:
Predicted instance:
https://img.onl/AEFAqj

After prediction, I corrected one of the nodes (posterior right leg):
https://img.onl/8ls4Px

I then moved to the next frame, and created a new instance by right clicking and selecting "Copy Prior Frame"
https://img.onl/R7rsE
Here, the posterior right leg is in the uncorrected position. 
My expectation was that the instances on the last two figures would be identical.

roomrys:
Thanks for reporting @Deruijter,

A bit of strange logic here (I'll need to double check why this is the case)... It seems SLEAP checks if the number of instances in the previous frame is greater than the number of instances in the current frame and uses the previous frame instance at index equal to the number of instances in the current frame.
https://github.com/talmolab/sleap/blob/19b95037cd56ffcc367bd16e6b949c2640aa466f/sleap/gui/commands.py#L2576-L2603

The last conditional would give us the result we want in this case (using the last instance added in the previous frame), but we never reach that bit of code.

Regardless, we should be preferring user instances over predicted instances or at least check whether the grabbed instance has an associated user instance.
https://github.com/talmolab/sleap/blob/19b95037cd56ffcc367bd16e6b949c2640aa466f/sleap/instance.py#L355-L364

Deruijter:
[...]

PR #1630, opened 2023-12-08
Resolve copy prior frame issue to take latest available user instance…

vaibhavtrip29:
… in prior frame rather than take the last instance in previous frame

### Description
Instead of returning the copy_instance, first check if the copy_instance is a predicted instance in which case check the previous frame to see if it has any user instances. If so, set copy_instance to the last user instance in the previous frame.

### Types of changes

- [X] Bugfix
- [ ] New feature
- [X] Refactor / Code style update (no logical changes)
- [ ] Build / CI changes
- [ ] Documentation Update
- [ ] Other (explain)

[list open issues here]
https://github.com/talmolab/sleap/issues/1065

:heart:

## Summary by CodeRabbit

- **Refactor**
  - Enhanced the logic for selecting instances to copy from, ensuring accurate handling of predicted and user instances.
- **Tests**
  - Added a new test `test_CopyPriorFrame` to validate the behavior of adding a user instance based on the previous frame's data.
- **Documentation**
  - Updated the declarations of exported or public entities for clarity.

### DIFF
--- a/sleap/gui/commands.py
+++ b/sleap/gui/commands.py
@@ -3067,15 +3067,25 @@ class AddInstance(EditCommand):
             prev_idx = AddInstance.get_previous_frame_index(context)
 
             if prev_idx is not None:
-                prev_instances = context.labels.find(
+                prev_frame = context.labels.find(
                     context.state["video"], prev_idx, return_new=True
-                )[0].instances
+                )[0]
+                prev_instances = prev_frame.instances
                 if len(prev_instances) > len(context.state["labeled_frame"].instances):
                     # If more instances in previous frame than current, then use the
                     # first unmatched instance.
                     copy_instance = prev_instances[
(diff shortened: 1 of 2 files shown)


roomrys:
@coderabbitai review

roomrys (commented):
I think we should always do the check for user instances (in all three cases) - maybe we could move the logic out since it's done in all the cases? It wouldn't help with computational cost, but it would help with readability. Also, for the lint test, we use black to format the code. You can either [set-up black to run each time you save](https://marcobelo.medium.com/setting-up-python-black-on-visual-studio-code-5318eba4cd00) ([vscode link](https://code.visualstudio.com/docs/python/formatting) if medium is paywalled) a file or run `black <the name of the file>` or if you want to do it more automated but not set it up in vscode:
```
git diff --name-only | grep . | xargs black
```

PR #1836, opened 2024-06-27
Use oks score for reprojections

roomrys:
### Description

This PR does 3 main things (described in more detail further down):
1. Use same OKS score for all `PredictedInstance.score`s
2. Add `InstanceGroup.score` attribute
3. Display `InstanceGroup.score` in table

Additionally, to use the `sleap.nn.evals.compute_oks` function inside a `sleap.io` file, we needed to move the `compute_oks` function to `sleap.util` to avoid a circular import between `sleap.nn` and `sleap.io` files.

#### Use same OKS score for all `PredictedInstance.score`s
Previously, all the `PredictedInstances` created and updated through (re)projection had a default nan score. This PR uses an average oks (calculated from ground truth `Instance`s and their reprojections) as the score for all (re)projected `PredictedInstance`s.

#### Add `InstanceGroup.score` attribute
To be most correct, even the `Instance`s which have reprojected nodes should have a score associated with them. There is currently no `Instance.score` attribute, and since this score really applies to the entire `InstanceGroup` (not just the `PredictedInstance`s in the `InstanceGroup`), this PR also adds an `InstanceGroup.score` attribute that is updated each time we call `InstanceGroup.update_points`.

#### Display `InstanceGroup.score` in table
This `InstanceGroup.score` is likely the most useful information for users to see, so this PR also adds a score column to the table displaying InstanceGroup information.

### Types of changes

- [ ] Bugfix
- [x] New feature
- [ ] Build / CI changes
- [ ] Documentation Update
- [ ] Other (explain)

[list open issues here]

:heart:


sleap/io/cameras.py
```diff
@@ -1676,19 +1746,17 @@ def numpy(
                         f"{self.instance_groups}"
                     )
 
-        instance_group_numpys: List[np.ndarray] = []  # len(T) M=all x N x 2
+        instance_group_numpys: List[np.ndarray] = []  # len(T) M=include x N x 2
```
eberrigan, line 1749: is the comment right here?

tests/io/test_cameras.py
```diff
@@ -590,6 +606,20 @@ def test_instance_group(
...
+        score=0.5,
+    )
+    assert instance_group_from_dict.score == 0.5
+    # The score of instances will NOT be updated on initialization.
+    for instance in instance_group_from_dict.instances:
+        if isinstance(instance, PredictedInstance):
+            assert instance.score != instance_group_from_dict.score
 
     # Test `__repr__`
     print(instance_group)
```
eberrigan, line 625: did you mean to leave this?

eberrigan (approved):
Looks good! Just made some comments about a print statement and cleaning up a comment

tests/io/test_cameras.py
roomrys: Yes, the `__repr__` method is used when calling `print`

sleap/io/cameras.py
roomrys: Yes, in the for loop below, we add `T` elements to the `instance_group_numpys` list each with shape `M=include x N x 2`

PR #1816, opened 2024-06-20
Create new instances away from copied instances

Arthurmayo:
### Description
A new instance is created 10 pixels away from the instance it is copying from. This allows for faster labeling. 

### Types of changes

- [ ] Bugfix
- [x] New feature
- [ ] Refactor / Code style update (no logical changes)
- [ ] Build / CI changes
- [ ] Documentation Update
- [ ] Other (explain)

### Does this address any currently open issues?
[list open issues here]

### Outside contributors checklist

- [ ] Review the [guidelines for contributing](https://github.com/talmolab/sleap/blob/develop/docs/CONTRIBUTING.md) to this repository
- [ ] Read and sign the [CLA](https://github.com/talmolab/sleap/blob/develop/sleap-cla.pdf) and add yourself to the [authors list](https://github.com/talmolab/sleap/blob/develop/AUTHORS)
- [ ] Make sure you are making a pull request against the **develop** branch (not *main*). Also you should start *your branch* off *develop*
- [ ] Add tests that prove your fix is effective or that your feature works
- [ ] Add necessary documentation (if appropriate)

#### Thank you for contributing to SLEAP!
:heart:

## Summary by CodeRabbit

- **Bug Fixes**
  - Adjusted the positioning logic for predicted instances to ensure accurate x and y coordinates.
  - Updated video widget logic to make all nodes selectable, enhancing usability and consistency.

PR #1859, opened 2024-07-10
Added Three Different Cases for Adding a New Instance

7174Andy:
### Description
1. Ctrl + I should create instances slightly away from the copied instance
2. Ctrl + V should create instances exactly on top of the instances it's copying from
3. Right click and click "Default" to create instance should make the instance based on the location of the click

### Types of changes

- [ ] Bugfix
- [x] New feature
- [ ] Build / CI changes
- [ ] Documentation Update
- [ ] Other (explain)

This PR is an expansion of PR #1816. 

- [x] Review the [guidelines for contributing](https://github.com/talmolab/sleap/blob/develop/docs/CONTRIBUTING.md) to this repository
- [x] Read and sign the [CLA](https://github.com/talmolab/sleap/blob/develop/sleap-cla.pdf) and add yourself to the [authors list](https://github.com/talmolab/sleap/blob/develop/AUTHORS)
- [x] Add necessary documentation (if appropriate)

:heart:

## Summary by CodeRabbit

- **New Features**
  - Introduced the ability to specify a `location` for node positions, enhancing control over instance initialization and visibility settings.
  - Updated the contextual menu in the video widget to support initializing new instances with a specific `location`.


sleap/gui/commands.py
```diff
@@ -2984,6 +2985,7 @@ def set_visible_nodes(
         copy_instance: Optional[Union[Instance, PredictedInstance]],
         new_instance: Instance,
         mark_complete: bool,
+        location: Optional[QtCore.QPoint],
```
roomrys, line 2988: When the type can be optional, I like to add a default value of `None` so that we can call this method without needing to specify the location:
```suggestion
        location: Optional[QtCore.QPoint] = None,
```

sleap/gui/commands.py
```diff
@@ -3018,8 +3026,29 @@ def set_visible_nodes(
...
                 x_old = copy_instance[node].x
                 y_old = copy_instance[node].y
-                x_new = x_old * scale_width
-                y_new = y_old * scale_height
+
+                # Handles if the new instance is created with right click
+                if (
+                    location is not None
+                    and offset_x is not None
+                    and offset_y is not None
```
roomrys, line 3034: Do we ever expect the offsets to be `None`? I don't believe we initialize either `offset_x` or `offset_y` and they are only ever defined if `location is not None` (in which case I expect neither `offset_x` nor `offset_y` to be `None`)

sleap/gui/commands.py
```diff
@@ -3018,8 +3026,29 @@ def set_visible_nodes(
...
+                ):
+                    x_old += offset_x
+                    y_old += offset_y
```
roomrys, line 3037: If you apply the offset here, then you will be scaling the offset below resulting in the instance being offset a scaled offset amount which is undesirable.

sleap/gui/commands.py
```diff
@@ -3018,8 +3026,29 @@ def set_visible_nodes(
...
+                else:
+                    if (x_old + 10) * scale_width <= (new_size_width - 10):
+                        x_new = (x_old + 10) * scale_width
+                    else:
+                        x_new = x_old
+
+                    if (y_old + 10) * scale_height <= (new_size_height - 10):
+                        y_new = (y_old + 10) * scale_height
+                    else:
+                        y_new = y_old
```
roomrys, line 3051: Add a comment to each conditional block for easier reading, ie.:
```python
# Copy the instance without scale or offset if predicted
if isinstance(copy_instance, PredictedInstance):
```

roomrys (changes requested):
Have not gotten a chance to test yet, but some comments. Most important is how the offset will be affected when scaling a few lines below.

sleap/gui/commands.py
```diff
@@ -2984,6 +2985,7 @@ def set_visible_nodes(
         copy_instance: Optional[Union[Instance, PredictedInstance]],
         new_instance: Instance,
         mark_complete: bool,
```
roomrys, line 2987: We might also want to pass in the `init_method` into `AddInstance.set_visible_nodes` and use `init_method == "best"` in addition to `location is not None` to determine whether the offset should be non-zero.

roomrys (changes requested):
The current implementation will add (a scaled) +10 offset even if we are in the third case
> 3. Right click and click "Default" to create instance should make the instance based on the location of the click

or the unmentioned fourth case
> 4. Neither cases 1-3

We want to add an unscaled +10 offset on in the first case:
>1. Ctrl + I should create instances slightly away from the copied instance

And also handle the fourth case better (i.e. 0 offset). See suggestions/comments

sleap/gui/commands.py
```diff
@@ -3010,6 +3014,14 @@ def set_visible_nodes(
         scale_width = new_size_width / old_size_width
         scale_height = new_size_height / old_size_height
 
+        # Calculate the difference from the cursor to the
+        # original position for getting new x and y values.
+        if location is not None and init_method == "best":
+            reference_x = copy_instance[context.state["skeleton"].node_names[0]].x
+            reference_y = copy_instance[context.state["skeleton"].node_names[0]].y
+            offset_x = location.x() - reference_x
+            offset_y = location.y() - reference_y
```
roomrys, line 3023: Also, the reference_x and y here should be the scaled reference.

sleap/gui/commands.py
```diff
@@ -3010,6 +3014,25 @@ def set_visible_nodes(
...
+        # Using the menu or the hotkey
+        if init_method == "best":
+            offset_x = 10
+            offset_y = 10
+
+            # Using right click and context menu
+            if location is not None:
+                reference_node = next(
+                    (node for node in copy_instance if not node.isnan()), None
+                )
```
roomrys, line 3030: Finds first non nan reference node

sleap/gui/commands.py
```diff
@@ -3018,13 +3041,45 @@ def set_visible_nodes(
...
+                else:
+                    x_new = x_new_offset
+                if y_new_offset < 0:
+                    y_new = 0
+                    visible = False
+                elif y_new_offset > new_size_height:
+                    y_new = new_size_height
+                    visible = False
+                else:
+                    y_new = y_new_offset
```
roomrys, line 3076: Moves out of bounds nodes to closest in bounds position. Also marks those nodes that were going to be out of bounds as not visible.

roomrys (commented):
I pushed a few changes. Do you mind reviewing? The thing I am most uncertain about is setting the OOB nodes to not visible after shifting them in bounds.

roomrys (approved):

PR #1860, opened 2024-07-10
Add highlight for instance group dock

justinvshen:
### Description
add highlighting feature for row in instance group dock

### Types of changes

- [ ] Bugfix
- [X] New feature
- [ ] Build / CI changes
- [ ] Documentation Update
- [ ] Other (explain)

[list open issues here]

:heart:

### DIFF
--- a/sleap/gui/widgets/video.py
+++ b/sleap/gui/widgets/video.py
@@ -70,7 +70,6 @@ from sleap.gui.shortcuts import Shortcuts
 from sleap.gui.state import GuiState
 from sleap.gui.widgets.slider import VideoSlider
 from sleap.instance import Instance, Point, PredictedInstance
-from sleap.io.cameras import InstanceGroup
 from sleap.io.video import Video
 from sleap.prefs import prefs
 from sleap.skeleton import Node
@@ -962,7 +961,9 @@ class GraphicsView(QGraphicsView):
         scene_items = self.scene.items(Qt.SortOrder.AscendingOrder)
         return list(filter(lambda x: isinstance(x, QtInstance), scene_items))
 
-    def selectInstance(self, select: Union[Instance, int, InstanceGroup]):
+    
+
+    def selectInstance(self, select: Union[Instance, int, 'QtInstance']):
         """
         Select a particular instance in view.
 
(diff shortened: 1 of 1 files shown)


sleap/gui/widgets/video.py
```diff
@@ -970,8 +972,16 @@ def selectInstance(self, select: Union[Instance, int]):
...
             None
         """
+        
         for idx, instance in enumerate(self.all_instances):
+            if isinstance(select, int):
+                instance.selected = select == idx
+            elif isinstance(select, Instance):
+                instance.selected = select == instance.instance
+            elif isinstance(select, InstanceGroup):
+                instance.selected = True if instance in select else False
```
roomrys, line 982: `InstanceGroup` stores `Instance` objects not `QtInstance` objects, so we need to use the `QtInstance.instance` attribute to get the associated `Instance`. 

Also, when using the `in` keyword in Python, the dunder method `__contains__` is called on the object that is being checked for containment. If `__contains__` is not implemented, Python falls back to using the `__iter__` method to iterate over the elements (if available) and checks each element. If `__iter__` is also not implemented, Python then tries to use the `__getitem__` method to check for containment by iterating over index values starting from zero.

Neither `InstanceGroup.__contains__` not `InstanceGroup.__iter__` is implmented and `InstanceGroup.__getitem__` unfortunately only handles this in the case of the `Instance` selected being the first instance in `InstanceGroup.instances` (we iterate to too large a number otherwise and get a `KeyError`). Let's just use `in InstanceGroup.instances` for now.
```suggestion
                instance.selected = True if instance.instance in select.instances else False
```

roomrys (changes requested):
If we make these changes, then things should be in a working state. Also, make sure to run `black` on `sleap/gui/widgets/video.py` (and any other file you change - if any)

sleap/gui/widgets/video.py
```diff
@@ -970,8 +973,15 @@ def selectInstance(self, select: Union[Instance, int]):
...
             None
         """
+        
         for idx, instance in enumerate(self.all_instances):
-            instance.selected = select == idx or select == instance.instance
+            if isinstance(select, int):
+                instance.selected = select == idx
+            elif isinstance(select, Instance):
+                instance.selected = select == instance.instance
+            elif isinstance(select, QtInstance.instance):
```
roomrys, line 982: We cannot access an instance attribute on a class. If you have some time, I would suggest reading this [article on instance attributes](https://realpython.com/python-classes/#instance-attributes).
```suggestion
            elif isinstance(select, QtInstance):
```

roomrys (changes requested):
Bug has been located. Sorry if my comments mislead you!

PR #1862, opened 2024-07-10
Generate suggestions using max point displacement threshold

gqcpm:
### Description
1. Created 2 new functions that should resolve the feature request #1761: function max_point_displacement, _max_point_displacement_video. 
2. Added to yaml file to include max_point_displacement. 
3. Create test for new function . . . will need to edit

### Types of changes

- [ ] Bugfix
- [x] New feature
- [ ] Build / CI changes
- [ ] Documentation Update
- [ ] Other (explain)

- #1761

:heart:

## Summary by CodeRabbit

- **New Features**
  - Introduced a new configuration option for `max point displacement` to help users detect frames with significant point movement.
  - Enhanced functionality for analyzing point displacement in video frames, allowing users to set a maximum displacement threshold.
  
- **Tests**
  - Added a new test for the `max point displacement` feature to ensure its accuracy and reliability.


sleap/config/suggestions.yaml
```diff
@@ -175,6 +175,12 @@ main:
         type: double
         default: 0.1
         range: 0.1,1.0
+    
+    "max point displacement":
+      - name: per_video
```
roomrys, line 180: Perhaps we should give this a more accurate/descriptive name such as:
```suggestion
      - name: threshold
```

sleap/gui/suggestions.py
```diff
@@ -291,6 +292,57 @@ def _velocity_video(
...
+        for i in range(1, frames):
+            prev_lf = lfs[i - 1]
+            curr_lf = lfs[i]
+            prev_points = np.array([inst.points_array for inst in prev_lf.instances_to_show])
+            curr_points = np.array([inst.points_array for inst in curr_lf.instances_to_show])
+
+            if prev_points.shape != curr_points.shape:
+                continue
+
+            displacement = np.linalg.norm(curr_points - prev_points, axis=2).sum()
```
roomrys, line 336: This calculation for displacement will not gaurantee that we compare the instances we want to because `LabeledFrame.instances_to_show` returns `Instance`s in the order that they were added to the `LabeledFrame` - i.e. there is no gaurantee that `prev_lf.instances_to_show[0]` is annotating the same animal as `curr_lf.instances_to_show[0]`.

Instead, we should be finding `Instance`s that are assigned to the same `Track` - which will guarantee that they are annotating the same animal (or at least attempting to annotate the same animal).

The `Labels.numpy` method returns us a numpy array for all `LabeledFrame`s and all `Instance`s sorted by `Track`:
https://github.com/talmolab/sleap/blob/14c21b4459773c99b941b77867d5746becc68ff7/sleap/io/dataset.py#L2442-L2468

roomrys (changes requested):
Make sure that you are comparing displacements for `Instance`s of the same `Track` (see review comments).

sleap/gui/suggestions.py
```diff
@@ -291,6 +293,86 @@ def _velocity_video(
...
+
+            if valid_prev_points.size == 0 or valid_curr_points.size == 0:
+                continue
+
+            displacement = np.linalg.norm(valid_curr_points - valid_prev_points, axis=1).sum()
+            displacements.append((displacement, idx))
+
+        frame_idxs = [
+            frame_idx for displacement, frame_idx in displacements if displacement > displacement_threshold
+        ]
```
roomrys, line 371: The suggested approach handles nan values as we want by:
1. resulting in nan in the euclidean norm
2. Being excluded in the mean calculation for all points in an `Instance`

```suggestion
        # Get numpy of shape (frames, tracks, nodes, x, y)
        labels_numpy = labels.numpy(video=video, all_frames=True, untracked=False)

        # Return empty list if not enough frames
        n_frames, n_tracks, n_nodes, _ = labels_numpy.shape
        if n_frames < 2:
            return []

        # Calculate displacements
        diff = labels_numpy[1:] - labels_numpy[:-1]  # (frames - 1, tracks, nodes, x, y)
        euc_norm = np.linalg.norm(diff, axis=-1)  # (frames - 1, tracks, nodes)
        mean_euc_norm = np.nanmean(euc_norm, axis=-1)  # (frames - 1, tracks)

        # Find frames where mean displacement is above threshold
        threshold_mask = np.any(
            mean_euc_norm > displacement_threshold, axis=-1
        )  # (frames - 1,)
        frame_idxs = list(np.argwhere(threshold_mask).flatten())  # [0, len(frames - 1)]
```

roomrys (changes requested):
Handle nans appropriately and get rid of unneeded loops (see comment).

roomrys (approved):

Issue #1918, opened 2024-08-26
Error during Inference. ValueError: Found Multiple nodes named (name).


hemanyaradadia:
# Bug description

Running Inference and Tracking throws up 'ValueError: Found multiple nodes named (name).' on a Linux system. While running the same command on a Windows 10 system, it works without any error.  

## Expected behaviour

A .slp output file with predictions/track in each frame. 

## Actual behaviour

The error shows up in the terminal. While .slp file is generated but without any predictions or frames. Essentially an empty .slp project file.  

## Your personal set up

- OS: Proxmox VE 8.1.4

- Version(s): SLEAP 1.3.3 Python 3.7.12

- SLEAP installation method (listed [here](https://sleap.ai/installation.html#)):
  - [X] [Conda from package](https://sleap.ai/installation.html#conda-package)
  - [ ] [Conda from source](https://sleap.ai/installation.html#conda-from-source)
  - [ ] [pip package](https://sleap.ai/installation.html#pip-package)
  - [ ] [Apple Silicon Macs](https://sleap.ai/installation.html#apple-silicon-macs)

<details><summary>Environment packages</summary>

```
# packages in environment at /mnt/pve/Homes/hemanya/miniconda3/envs/sleap-env:
#
# Name                    Version                   Build  Channel
_libgcc_mutex             0.1                 conda_forge    conda-forge
_openmp_mutex             4.5                       2_gnu    conda-forge
absl-py                   1.0.0                    pypi_0    pypi
aiofiles                  22.1.0             pyhd8ed1ab_0    conda-forge
aiosqlite                 0.20.0             pyhd8ed1ab_0    conda-forge
alsa-lib                  1.2.3.2              h166bdaf_0    conda-forge
anyio                     3.7.1              pyhd8ed1ab_0    conda-forge
argon2-cffi               23.1.0             pyhd8ed1ab_0    conda-forge
argon2-cffi-bindings      21.2.0           py37h540881e_2    conda-forge
arrow                     1.2.3              pyhd8ed1ab_0    conda-forge
astunparse                1.6.3                    pypi_0    pypi
attrs                     21.4.0             pyhd8ed1ab_0    conda-forge
babel                     2.14.0             pyhd8ed1ab_0    conda-forge
backcall                  0.2.0              pyh9f0ad1d_0    conda-forge
backports                 1.0                pyhd8ed1ab_4    conda-forge
backports-zoneinfo        0.2.1                    pypi_0    pypi
[...]

PR #1934, opened 2024-09-03
Refactor `skeleton.to_json` to serialize skeleton object without jsonpickle

eberrigan:
### Description
- We are seeing issues with `skeleton.json` where `jsonpickle` is encoding the skeleton graph dictionary in a way that breaks backwards compatibility, specifically:
     - """The change in attrs version used by jsonpickle to write/read the skeleton.json is causing problems, namely, attrs >=22.2.0 returns the object state as a dictionary instead of a tuple. This results in incompatibilities when attempting to deserialize skeleton.json files (w/ an older attrs <22.2.0) that were serialized with attrs >=22.2.0. Based on the discussion in https://github.com/python-attrs/attrs/pull/1085 it sounds like attrs ==22.2.0 was unable to deserialize pickled data from attrs <22.2.0, and attrs ==23.1.0 (>=23.1.0?) at least fixes backwards compatibility. The long-term plan for attrs (discussed in https://github.com/python-attrs/attrs/issues/1091) is to remove support for deserializing tuples after giving everyone a few versions of attrs to convert their data over to a dictionary serialization.""" from #1918 .
     - https://github.com/talmolab/sleap/blob/1370782877edc10346023fecbaf502e5bf3ce006/sleap/skeleton.py#L986-L1028
- The solution is to move away from `jsonpickle` and manually serialize our skeleton objects to json strings.
- We can re-use code written in `sleap-io` to do this
     - https://github.com/talmolab/sleap-io/blob/2bc3d5210c46bdb25413d25970c4bdc7adb6e8cc/sleap_io/io/slp.py#L606-L710
- Tests will be added to make sure the new serialized skeletons can be used as expected
     - https://github.com/talmolab/sleap/blob/1370782877edc10346023fecbaf502e5bf3ce006/sleap/skeleton.py#L1057 
### Types of changes

- [ ] Bugfix
- [ ] New feature
- [X] Refactor / Code style update (no logical changes)
- [ ] Build / CI changes
- [ ] Documentation Update
- [ ] Other (explain)

- #1918 
- #1841

:heart:

## Summary by CodeRabbit

- **Chores**
	- Added a TODO comment for future enhancement regarding JSON encoding in the Skeleton class.
- **New Features**
	- Improved the `to_json` method for the Skeleton class, enhancing the serialization of skeleton data with a more structured representation of nodes and edges.
	- Introduced a new Jupyter notebook for testing the conversion of skeleton data to JSON format, demonstrating serialization and deserialization processes.

### DIFF
--- a/sleap/skeleton.py
+++ b/sleap/skeleton.py
@@ -1050,14 +1050,8 @@ class Skeleton:
                     # in legacy SLEAP.
                     "edge_insert_idx": edge_ind,
                     "key": 0,  # Always 0.
-                    "source": {
-                        "py/object": "sleap.skeleton.Node",
-                        "py/state": {"py/tuple": [source.name, source.weight]},
-                    },
-                    "target": {
-                        "py/object": "sleap.skeleton.Node",
-                        "py/state": {"py/tuple": [target.name, target.weight]},
-                    },
+                    "source": {"py/id": self.nodes.index(node)},
+                    "target": {"py/id": self.nodes.index(node)},
                     "type": edge_type,
                 }
             )
--- a/sleap/skeleton.py
+++ b/sleap/skeleton.py
@@ -999,6 +999,7 @@ class Skeleton:
         Returns:
             A string containing the JSON representation of the skeleton.
         """
+        # TODO: Replace jsonpickle with a custom encoder from sleap-io.
         jsonpickle.set_encoder_options("simplejson", sort_keys=True, indent=4)
         if node_to_idx is not None:
             indexed_node_graph = nx.relabel_nodes(
--- a/sleap/skeleton.py
+++ b/sleap/skeleton.py
@@ -1078,14 +1078,14 @@ class Skeleton:
             else:
                 edge_type = {"py/id": 2}
 
-            src, dst = tuple(symmetry.nodes)
-            print(f"src: {src}")
-            print(f"dst: {dst}")
+            symmetry_src, symmetry_dst = symmetry
+            print(f"src: {symmetry_src}")
+            print(f"dst: {symmetry_dst}")
             edges_dicts.append(
                 {
                     "key": 0,
-                    "source": {"py/id": node_to_id[src]},
-                    "target": {"py/id": node_to_id[dst]},
+                    "source": {"py/id": node_to_id[symmetry_src]},
+                    "target": {"py/id": node_to_id[symmetry_dst]},
                     "type": edge_type,
                 }
             )
--- a/sleap/skeleton.py
+++ b/sleap/skeleton.py
@@ -1097,7 +1097,7 @@ class Skeleton:
                     {
                         "id": {
                             "py/object": "sleap.skeleton.Node",
-                            "py/state": {"name": node.name, "weight": node.weight},
+                            "py/state": {"py/tuple": [node.name, node.weight]},
                         }
                     }
                     for node in self.nodes
@@ -1118,7 +1118,7 @@ class Skeleton:
                     {
                         "id": {
                             "py/object": "sleap.skeleton.Node",
-                            "py/state": {"name": node.name, "weight": node.weight},
+                            "py/state": {"py/tuple": [node.name, node.weight]},
                         }
                     }
                     for node in self.nodes
--- a/sleap/skeleton.py
+++ b/sleap/skeleton.py
@@ -1082,15 +1082,16 @@ class Skeleton:
                     "type": edge_type,
                 }
             )
-
+        # Create graph field
+        graph = {
+            "name": self.name,
+            "num_edges_inserted": len(self.edges),
+        }
         # Create skeleton dict.
         if self.is_template:
             skeleton_dict = {
                 "directed": True,
-                "graph": {
-                    "name": self.name,
-                    "num_edges_inserted": len(self.edges),
-                },
+                "nx_graph": graph, 
                 "links": edges_dicts,
                 "multigraph": True,
                 # In the order in Skeleton.nodes and must match up with nodes_dicts.
@@ -1101,14 +1102,16 @@ class Skeleton:
         else:
             skeleton_dict ={
                     "directed": True,
-                    "graph": {
-                        "name": self.name,
-                        "num_edges_inserted": len(self.edges),
-                    },
+                    "nx_graph": graph,
                     "links": edges_dicts,
                     "multigraph": True,
                     # In the order in Skeleton.nodes and must match up with nodes_dicts.
                     "nodes": [{"id": {"py/id": node_to_id[node]}} for node in self.nodes],}
+            
+        print(f'skeleton_dict: {skeleton_dict}')
+        json_str = json.dumps(skeleton_dict)
+        print(f'json_str: {json_str}')
+        return json_str
 
         # jsonpickle.set_encoder_options("simplejson", sort_keys=True, indent=4)
         # if node_to_idx is not None:
--- a/sleap/skeleton.py
+++ b/sleap/skeleton.py
@@ -1005,13 +1005,17 @@ class Skeleton:
         nodes_dicts = []
         node_to_id = {}
         for node in self.nodes:
+            print(f'node: {node}')
             if node not in node_to_id:
                 node_to_id[node] = node_to_idx[node] if node_to_idx is not None else len(node_to_id)
                 nodes_dicts.append({"name": node.name, "weight": 1.0})
+                print(f'node_to_id: {node_to_id}')
+                print(f'nodes_dicts: {nodes_dicts}')
         
         # Create a dictionary to store edge data
         edges_dicts = []
         for edge_ind, edge in enumerate(self.edges):
+            print(f'edge: {edge}')
             if edge_ind == 0:
                 edge_type = {
                     "py/reduce": [
@@ -1019,19 +1023,20 @@ class Skeleton:
                         {"py/tuple": [1]},  # 1 = real edge, 2 = symmetry edge
                     ]
                 }
+                print(f'edge_type: {edge_type}')
             else:
                 edge_type = {"py/id": 1}
-
+                print(f'edge_type: {edge_type}')
             edges_dicts.append(
                 {
                     "edge_insert_idx": edge_ind,
                     "key": 0,  # Always 0.
-                    "source": node_to_id[edge.source],
-                    "target": node_to_id[edge.destination],
+                    "source": node_to_id[edge[0]],
+                    "target": node_to_id[edge[1]],
                     "type": edge_type,
                 }
             )
-
+            print(f'edges_dicts: {edges_dicts}')
         # Build links dicts for symmetry edges.
         for symmetry_ind, symmetry in enumerate(self.symmetries):
             if symmetry_ind == 0:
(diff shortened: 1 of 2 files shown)


talmo:
Relevant sleap-io portion:

https://github.com/talmolab/sleap-io/blob/2bc3d5210c46bdb25413d25970c4bdc7adb6e8cc/sleap_io/io/slp.py#L606-L710

^ That is 100% self-contained and can probably be ported over directly.

A nicer (if more onerous) approach would be to finally include sleap-io as a dependency and just deep import that exact function and nothing else (for this PR).

For example, in SLEAP's `Skeleton.to_json` method, you could do something like:

```py
def to_json(self, ...):
    from sleap_io.io.slp import serialize_skeletons

    skeleton_dicts = serialize_skeletons([self])

    ...
```

This should probably happen at a higher level though and just refactor the serialization logic.

Also note a couple of gotchas with the sleap-io implementation:
- The insert idx is not exactly the same [[1]](https://github.com/talmolab/sleap-io/blob/2bc3d5210c46bdb25413d25970c4bdc7adb6e8cc/sleap_io/io/slp.py#L635-L642)[[2]](https://github.com/talmolab/sleap-io/blob/2bc3d5210c46bdb25413d25970c4bdc7adb6e8cc/sleap_io/io/slp.py#L663-L664). This is generated by networkx I believe, which we don't use in sleap-io since we do all the serialization ourselves. Shouldn't matter functionally I think.
- We don't use the weights attribute. [[1]](https://github.com/talmolab/sleap-io/blob/2bc3d5210c46bdb25413d25970c4bdc7adb6e8cc/sleap_io/io/slp.py#L642)

eberrigan:
> Relevant sleap-io portion:
> 
> https://github.com/talmolab/sleap-io/blob/2bc3d5210c46bdb25413d25970c4bdc7adb6e8cc/sleap_io/io/slp.py#L606-L710
> 
> ^ That is 100% self-contained and can probably be ported over directly.
> 
> A nicer (if more onerous) approach would be to finally include sleap-io as a dependency and just deep import that exact function and nothing else (for this PR).
> 
> For example, in SLEAP's `Skeleton.to_json` method, you could do something like:
> 
> ```python
> def to_json(self, ...):
>     from sleap_io.io.slp import serialize_skeletons
> 
>     skeleton_dicts = serialize_skeletons([self])
> 
>     ...
> ```
> 
> This should probably happen at a higher level though and just refactor the serialization logic.
> 
> Also note a couple of gotchas with the sleap-io implementation:
> 
> * The insert idx is not exactly the same [[1]](https://github.com/talmolab/sleap-io/blob/2bc3d5210c46bdb25413d25970c4bdc7adb6e8cc/sleap_io/io/slp.py#L635-L642)[[2]](https://github.com/talmolab/sleap-io/blob/2bc3d5210c46bdb25413d25970c4bdc7adb6e8cc/sleap_io/io/slp.py#L663-L664). This is generated by networkx I believe, which we don't use in sleap-io since we do all the serialization ourselves. Shouldn't matter functionally I think.
> * We don't use the weights attribute. [[1]](https://github.com/talmolab/sleap-io/blob/2bc3d5210c46bdb25413d25970c4bdc7adb6e8cc/sleap_io/io/slp.py#L642)

- I can include the `networkx` indexing as we did before. 
- What do you mean by we don't use the weight attribute? I see it here but stored in the tuple: https://github.com/talmolab/sleap/issues/1918#issuecomment-2310870699.

sleap/skeleton.py
```diff
@@ -999,34 +999,153 @@ def to_json(self, node_to_idx: Optional[Dict[Node, int]] = None) -> str:
...
-            data = {
-                "nx_graph": graph,
+            # Template skeletons have additional fields
+            nx_graph = {
+                "directed": True,
+                "graph": graph, 
+                "links": edges_dicts,
+                "multigraph": True,
+                # In the order in Skeleton.nodes and must match up with nodes_dicts.
+                "nodes": [{"id": {"py/id": node_to_id[node]}} for node in self.nodes],
```
roomrys, line 1100: Assuming jsonpickle is decoding by reading things in order, we can assume that all `Node`s will only appear in the `"nodes"` list once. Therefore, we can (hopefully) move the `"nodes"` list before the `"links"` dictionary. The `"nodes"` list would need to define the nodes in the `{py/object: ..., py/state: ...}` dictionary. Then, the `"links"` dictionary should just reference nodes by `"py/id"` (referencing the `Node`'s index in the `"nodes"` list).
```suggestion
                "nodes": [{"id": {"py/object": "sleap.skeleton.Node", "py/state": {"name": node.name, "weight": node.weight}}} for node in self.nodes],
                "links": edges_dicts,
                "multigraph": True,
```

sleap/skeleton.py
```diff
@@ -999,34 +999,153 @@ def to_json(self, node_to_idx: Optional[Dict[Node, int]] = None) -> str:
...
-
-        json_str = jsonpickle.encode(data)
-
+            skeleton_dict ={
+                    "directed": True,
+                    "graph": graph,
+                    "links": edges_dicts,
+                    "multigraph": True,
+                    # In the order in Skeleton.nodes and must match up with nodes_dicts.
+                    "nodes": [{"id": {"py/id": node_to_id[node]}} for node in self.nodes],}
```
roomrys, line 1114: Assuming jsonpickle is decoding by reading things in order, we can assume that all `Node`s will only appear in the `"nodes"` list once. Therefore, we can (hopefully) move the `"nodes"` list before the `"links"` dictionary. The `"nodes"` list would need to define the nodes in the `{py/object: ..., py/state: ...}` dictionary. Then, the `"links"` dictionary should just reference nodes by `"py/id"` (referencing the `Node`'s index in the `"nodes"` list).
```suggestion
                    "nodes": [{"id": {"py/object": "sleap.skeleton.Node", "py/state": {"name": node.name, "weight": node.weight}}} for node in self.nodes],
                    "links": edges_dicts,
                    "multigraph": True,
```

sleap/skeleton.py
```diff
@@ -999,34 +999,153 @@ def to_json(self, node_to_idx: Optional[Dict[Node, int]] = None) -> str:
...
+                {
+                    # Note: Insert idx is not the same as the edge index in the skeleton
+                    # in legacy SLEAP.
+                    "edge_insert_idx": edge_ind,
+                    "key": 0,  # Always 0.
+                    "source": {"py/object": "sleap.skeleton.Node", "py/state": {"name": source.name, "weight": 1.0}},
+                    "target": {"py/object": "sleap.skeleton.Node", "py/state": {"name": target.name, "weight": 1.0}},
+                    # "target": {"py/id": node_to_id[target]},
+                    "type": edge_type,
+                }
```
talmo, line 1057: What about doing something like:

```suggestion
            if source not in node_to_id:
                source_dict = {"py/object": "sleap.skeleton.Node", "py/state": {"name": source.name, "weight": 1.0}}
                node_to_id[source] = len(node_to_id)
            else:
                source_dict = {"py/id": node_to_id[source]}
                
            if target not in node_to_id:
                target_dict = {"py/object": "sleap.skeleton.Node", "py/state": {"name": target.name, "weight": 1.0}}
                node_to_id[target] = len(node_to_id)
            else:
                target_dict = {"py/id": node_to_id[target]}
                
            edges_dicts.append(
                {
                    # Note: Insert idx is not the same as the edge index in the skeleton
                    # in legacy SLEAP.
                    "edge_insert_idx": edge_ind,
                    "key": 0,  # Always 0.
                    "source": source_dict,
                    "target": target_dict,
                    "type": edge_type,
                }
```

sleap/skeleton.py
```diff
@@ -999,34 +999,153 @@ def to_json(self, node_to_idx: Optional[Dict[Node, int]] = None) -> str:
...
+            print(f'target: {target}')
+            print(f'node_to_id[target]: {node_to_id[target]}')
+            edges_dicts.append(
+                {
+                    # Note: Insert idx is not the same as the edge index in the skeleton
+                    # in legacy SLEAP.
+                    "edge_insert_idx": edge_ind,
+                    "key": 0,  # Always 0.
+                    "source": {"py/object": "sleap.skeleton.Node", "py/state": {"name": source.name, "weight": 1.0}},
+                    "target": {"py/object": "sleap.skeleton.Node", "py/state": {"name": target.name, "weight": 1.0}},
```
roomrys, line 1054: If we are moving the `"nodes"` list above the `"links"` dictionary, then we should be able to just reference the `Node`'s `"py/id"` here (i.e. their index in that `"nodes"` list which is also the `Node`'s index in `Skeleton.nodes`)
```suggestion
                    "source": {"py/id": self.nodes.index(node)},
                    "target": {"py/id": self.nodes.index(node)},
```

roomrys (changes requested):
Let's try moving the `"nodes"` list above the `"links"` dictionary and see what happens

talmo:
From group discussion:

1. We want to unstructure in the legacy "tuple" format, even though attrs won't like it in the future so that we have broadest compatibility.
  - In addition to the proposed changes above, we need to make sure that the "global" nodes list in the `Labels` also serializes out to dictionaries in the tuple format. Specifically, this happens [here-ish in `Labels.to_json`](https://github.com/talmolab/sleap/blob/e4bb4449ee4907f8315ef9f64511a7aaa0c79155/sleap/io/dataset.py#L1935C22-L1935C52).
2. We will handle structuring ourselves, mimicking the logic in sleap-io. This way, we have more control over serialization format and when we switch to a replacement.
  - We will need to handle structuring in [`Skeleton.from_json` here](https://github.com/talmolab/sleap/blob/e4bb4449ee4907f8315ef9f64511a7aaa0c79155/sleap/skeleton.py#L1074-L1080).
  - This will require switching from `jsonpickle` to `simplejson` and then creating the objects ourselves by calling their constructors, manually parsing the metadata like we do in [`sleap-io` here](https://github.com/talmolab/sleap-io/blob/096eefd7bb9868588faad102be962e86b631e76a/sleap_io/io/slp.py#L540-L603) (note how we expect the old "tuple" format there as well).

---

## Notes on "tuple" vs "dict" format

When pickling an `attrs` object, the `__getstate__` method is used for pickling and first unstructures the `attrs` object into a dictionary. The format of this dictionary used to be tuple-based but is now dict-based in newer `attrs` (`>=22.2.0`) due to the discussion in https://github.com/python-attrs/attrs/issues/1004.

Current versions of `attrs` support the old format (which we use and in which our users have created many files that use it). At some point, this will no longer work as per https://github.com/python-attrs/attrs/issues/1091.

Our plan is to manually handle structuring/unstructuring rather than relying on `jsonpickle` or `attrs`, allowing us to define our own format and to support both legacy and new formats.

For reference, for a `Node` defined like:

```py
node = sleap.skeleton.Node(name="a", weight=1.0)
```

The **old format** ("tuple") unstructures this into dictionaries like:
```py
{"py/object": "sleap.skeleton.Node", "py/state": {"py/tuple": ["a", 1.0]}}
```

Whereas the **new format** ("dict") unstructures this into dictionaries like:
```py
{"py/object": "sleap.skeleton.Node", "py/state": {"name": "a", "weight": 1.0}}
```

roomrys:
Closed in favor of:
- #1970

PR #1939, opened 2024-09-04
Add `normalized_instance_similarity` method 

gitttt-1234:
### Description
Currently, we are facing [ID switches in tracking](https://github.com/talmolab/sleap/discussions/1815) because of very low similarity scores (~/= 0) with instance matching similarity function as it doesn't apply normalization. To address this issue, we add a new `normalized_instance_similarity` function which normalizes the keypoints based on the image size. 

### Types of changes

- [ ] Bugfix
- [ ] New feature
- [ ] Build / CI changes
- [ ] Documentation Update
- [ ] Other (explain)

#1815 

:heart:

## Summary by CodeRabbit

- **New Features**
	- Introduced a new similarity metric, **normalized_instance**, for enhanced tracking and analysis.
	- Added a new similarity metric, **object_keypoint**, for improved instance comparison.
	- Updated command-line interface options to include **normalized_instance** for tracking configurations.
	- Enhanced tracking functionality by incorporating image dimensions for more accurate processing.

- **Bug Fixes**
	- Improved the accuracy of instance matching by incorporating normalization in similarity assessments.

- **Tests**
	- Expanded test coverage to include the new **normalized_instance** and **object_keypoint** metrics for comprehensive validation of tracking functionality.

### DIFF
--- a/sleap/nn/inference.py
+++ b/sleap/nn/inference.py
@@ -2622,7 +2622,7 @@ class TopDownPredictor(Predictor):
                         # Set tracks for predicted instances in this frame.
                         predicted_instances = self.tracker.track(
                             untracked_instances=predicted_instances,
-                            img_hw=ex["image"].shape[-2:],
+                            img_hw=ex["image"].shape[-3:-1],
                             img=image,
                             t=frame_ind,
                         )
@@ -3265,7 +3265,7 @@ class BottomUpPredictor(Predictor):
                         # Set tracks for predicted instances in this frame.
                         predicted_instances = self.tracker.track(
                             untracked_instances=predicted_instances,
-                            img_hw=ex["image"].shape[-2:],
+                            img_hw=ex["image"].shape[-3:-1],
                             img=image,
                             t=frame_ind,
                         )
--- a/sleap/nn/inference.py
+++ b/sleap/nn/inference.py
@@ -2622,6 +2622,7 @@ class TopDownPredictor(Predictor):
                         # Set tracks for predicted instances in this frame.
                         predicted_instances = self.tracker.track(
                             untracked_instances=predicted_instances,
+                            img_hw=ex["image"].shape[-2:],
                             img=image,
                             t=frame_ind,
                         )
@@ -3264,6 +3265,7 @@ class BottomUpPredictor(Predictor):
                         # Set tracks for predicted instances in this frame.
                         predicted_instances = self.tracker.track(
                             untracked_instances=predicted_instances,
+                            img_hw=ex["image"].shape[-2:],
                             img=image,
                             t=frame_ind,
                         )
--- a/docs/guides/cli.md
+++ b/docs/guides/cli.md
@@ -207,7 +207,7 @@ optional arguments:
   --tracking.clean_iou_threshold TRACKING.CLEAN_IOU_THRESHOLD
                         IOU to use when culling instances *after* tracking. (default: 0)
   --tracking.similarity TRACKING.SIMILARITY
-                        Options: instance, centroid, iou (default: instance)
+                        Options: instance, normalized_instance, centroid, iou (default: instance)
   --tracking.match TRACKING.MATCH
                         Options: hungarian, greedy (default: greedy)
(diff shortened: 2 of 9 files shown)


getzze:
I'm just seeing this PR and it seems similar in spirit to the "object keypoint" similarity (https://github.com/talmolab/sleap/pull/1003) that was recently merged, but not correctly documented I'm afraid (I am the author so I am to blame!).

talmo:
> I'm just seeing this PR and it seems similar in spirit to the "object keypoint" similarity (#1003) that was recently merged, but not correctly documented I'm afraid (I am the author so I am to blame!).

Hi @getzze,

I think part of the problem is that we thought your method was not being hooked up to the GUI at all. Here is where we map the string names of the similarity methods to the actual functions:

https://github.com/talmolab/sleap/blob/e4bb4449ee4907f8315ef9f64511a7aaa0c79155/sleap/nn/tracking.py#L494-L499

The `object_keypoint` key is mapping to the same `instance_similarity`, though I see now upon closer inspection that it's getting special-cased here:

https://github.com/talmolab/sleap/blob/e4bb4449ee4907f8315ef9f64511a7aaa0c79155/sleap/nn/tracking.py#L886-L893

Docs would definitely help, but either way, I think what we're doing here is a bit different as it's basically just normalizing the keypoints by the image size rather than changing how we account for number of keypoints or confidence.

I suppose the best thing would be to have something more unified that can (optionally?) use all three factors.

For the moment, the easiest might just be to have separate methods as we'll be refactoring a lot of the tracking backend in the coming months anyway as part of our transition to PyTorch.

If you're curious or would like to weigh in on the design of the new backend, check out: https://github.com/talmolab/sleap-nn/issues/53 and the [current state of the implementation in this source tree](https://github.com/talmolab/sleap-nn/tree/58a9715e343effaf63bcf4fb3b5b765c7366279c/sleap_nn/tracking).

getzze:
> Docs would definitely help, but either way, I think what we're doing here is a bit different as it's basically just normalizing the keypoints by the image size rather than changing how we account for number of keypoints or confidence.

There is that, but the key point of `object_keypoint_similarity` is actually to normalize the distances before taking the exponential.

https://github.com/talmolab/sleap/blob/e4bb4449ee4907f8315ef9f64511a7aaa0c79155/sleap/nn/tracker/components.py#L123-L128

You can specify the normalization factor for each node, but what makes more sense is to use the standard error of infering the nodes (say 5 pixels). I didn't find a way to get this error simply though, so the extra parameter (but it's also more flexible like that).

Quoting what I wrote in the original PR #1003 :

> 1. Adding an scale to the distance between the reference and query keypoint. Otherwise, if the ref and query keypoints are 3 pixels apart, they contribute to 0.0001 to the similarity score, versus 0.36 if they are 1 pixel apart. This is very sensitive to single pixel fluctuations.
Instead, the distance is divided by a user-defined pixel scale before applying the gaussian function. The scale can be chosen to be the error for each keypoint found during training of the model with the validation set. Ideally this could be retrieved automatically, it is now hidden in the metrics.val.npz file of the model.
This is what they use in [this paper](https://arxiv.org/abs/1707.05388).

I should have put some explanation in the docs, really sorry about that...

So I think `instance_similarity`, `object_keypoint_similarity` and `normalized_instance_similarity` could be unified, with options to select one or the other.

Thanks for linking to the refactoring of tracking, the roadmap looks very exciting!

sleap/nn/inference.py
```diff
@@ -2622,6 +2622,7 @@ def _object_builder():
                         # Set tracks for predicted instances in this frame.
                         predicted_instances = self.tracker.track(
                             untracked_instances=predicted_instances,
+                            img_hw=ex["image"].shape[-3:-1],
```
roomrys, line 2625: From group review: Verify that channels are always last

roomrys (approved):

sleap/nn/inference.py
gitttt-1234: Verified!

In logs:
`INFO:sleap.nn.training:  Input shape: (192, 192, 1)`

PR #1988, opened 2024-10-09
Set selected instance to None when switching videos/frames

justinvshen:
### Description
Sets selected instance to None when switching videos/frames

### Types of changes

- [ ] Bugfix
- [X] New feature
- [ ] Build / CI changes
- [ ] Documentation Update
- [ ] Other (explain)

[list open issues here]

:heart:

### DIFF
--- a/sleap/gui/commands.py
+++ b/sleap/gui/commands.py
@@ -363,7 +363,7 @@ class CommandContext:
     def previousLabeledFrame(self):
         """Goes to labeled frame prior to current frame."""
         self.execute(GoPreviousLabeledFrame)
-        
+
     def nextLabeledFrame(self):
         """Goes to labeled frame after current frame."""
         self.execute(GoNextLabeledFrame)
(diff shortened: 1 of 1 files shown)


sleap/gui/commands.py
```diff
@@ -363,10 +363,12 @@ def exportFullPackage(self):
     def previousLabeledFrame(self):
         """Goes to labeled frame prior to current frame."""
         self.execute(GoPreviousLabeledFrame)
+        self.state["instance"] = None
```
roomrys, line 366: We want this logic to go inside the class that is being executed (i.e. `GoPreviousLabeledFrame`).

```suggestion
```

roomrys (changes requested):
Move `self.state["instance"]` logic into the classes that are being executed.

roomrys (changes requested):
Better, but I think we can simplify this further. Notice how most classes subclass either `NavCommand` or `GoIteratorCommand`. Therefore, we should be able to move the `context.state["instance"]` logic inside those classes to cover all cases.

roomrys:
After some manual testing, we realize that there are a few places in the GUI where we switch frames not through the `CommandContext`. Ideally, the `CommandContext` should handle all commands made by the user to the GUI. But, since we are on a divergent branch... we want to implement the smallest amount of code to fix the current problem (not properly resetting `state["instance"]` when switching frames). The minimal amount of code needed is a callback updating `state["instance"]` when we update `state["frame_idx"]`.

This PR is superseded by:
- #2012

PR #2028, opened 2024-11-20
Fix track only PredictedInstance

getzze:
### Description
Force to track only `PredictedInstance` and not `Instance`, otherwise we get an error because `update_matched_instance_tracks` only works `PredictedInstance`.
This can happen if running `tracking_only` with a dataset that has predicted instances from inference and user-defined instances used for training.

### Types of changes

- [x] Bugfix
- [ ] New feature
- [ ] Build / CI changes
- [ ] Documentation Update
- [ ] Other (explain)

[list open issues here]

:heart:


sleap/nn/tracking.py
```diff
@@ -1520,12 +1517,16 @@ def run_tracker(frames: List[LabeledFrame], tracker: BaseTracker) -> List[Labele
 
     # Run tracking on every frame
     for lf in frames:
+        # Use only the predicted instances
+        instances = [
+            inst for inst in lf.instances if isinstance(inst, PredictedInstance)
+        ]
```
talmo, line 1523: You can just use `lf.predicted_instances` for this

talmo:
Thanks @getzze! This looks like a useful filter, but we do sometimes have use cases where we want to reuse the user-labeled instances, for example, when correcting poses or tracking the same videos you label on.

If we can put this behind a feature flag so it's optional (and add some tests), happy to approve!

Thanks for the contribution!

getzze:
Thanks for reviewing! I will add an option to convert `Instances` to `PredictedInstances` before tracking.

getzze:
I made the proposed changes. One test is failing for no particular reason (it should pass if re-run).

PR #2511, opened 2025-12-23
Fix 'Delete All Predictions...' deleting user instances in multi-video projects

alicup29:
## Summary
  - Fixed bug where "Delete All Predictions..." menu item deleted user-labeled instances in multi-video projects
  - Root cause: `video.matches_content()` returns True for ALL videos with the same dimensions (shape and backend type), causing `remove_frames()` to remove frames from all videos instead of just the target
  - `remove_frames()` on videos w/ ImageVideo backend and 1 frame (`.tif`): frame to remove is idx 0 -> `remove_frames()` checks video shape & backend type + labeled frame idx (0 for .tif ImageVideos) -> call `labels_pop()`
  - soln: compare content AND path w/ [sleap-io `matches_path()` function](https://io.sleap.ai/latest/model/?h=matches_path#sleap_io.Video)

  ## Test plan
  - [ ] Open a multi-video project with both user instances and predictions
  - [ ] Use "Labels > Delete All Predictions..."
  - [ ] Verify only predictions are deleted, user instances are preserved
  - [ ] Verify frames with only user instances remain in the project

  🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/sleap/gui/commands.py
+++ b/sleap/gui/commands.py
@@ -118,6 +118,7 @@ from sleap.sleap_io_adaptors.lf_labels_utils import (
     remove_track,
     remove_video,
     remove_all_tracks,
+    videos_match,
 )
 
 # Indicates whether we support multiple project windows (i.e., "open" opens new window)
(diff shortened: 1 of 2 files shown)


gitttt-1234:
Fixed in the v1.6.0a3

PR #2750, opened 2026-06-08
Fix misleading error when running inference on empty project (#2749)

### LINKED ISSUE #2749: When clicking "run inference" on empty projects SLEAP raise wrong error
## Bug description

**Steps to recreate:**
1. run`sleap-label` on latest release
2. "Predict" > "Run Inference"
3. Error window pops up "You have unsaved changes. Please save before running training or inference"/ 

## Expected behaviour

It should raise an "No video selected"/"Project is empty"/etc. 

## Screenshots

Image

tom21100227:
## Description

Fixes #2749.

When running `sleap-label` on a fresh, empty project and clicking **Predict → Run Inference** (or training), SLEAP showed the wrong error:

> You have unsaved changes. Please save before running training or inference.

…even though an empty project has no changes.

### Root cause

The pre-flight guard in `_show_learning_dialog` (`sleap/gui/app.py`) collapsed two distinct conditions into a single message:

```python
if not self.state["filename"] or self.state["has_changes"]:
    QMessageBox(text="You have unsaved changes. Please save before running training or inference.").exec_()
    return
```

A brand-new project trips the `not filename` case (it was never saved) while `has_changes` is `False`, so the user is incorrectly told they have unsaved changes.

### Fix

Split the guard into three ordered checks, each with an accurate message:

1. **No videos** → "This project has no videos. Please add a video before running training or inference."
2. **Never saved** (`not filename`) → "Please save your project before running training or inference."
3. **Unsaved changes** (`has_changes`) → original "You have unsaved changes…" message.

The no-videos check comes first so a fresh empty project gets the most relevant message, matching the expected behavior in the issue.

The underlying design is unchanged: saving is still required before running, since inference runs as a subprocess that reads the `.slp`/video from disk via `--data_path`. Only the messaging was corrected.

## Types of changes

- [x] Bugfix

## Testing

Added `test_show_learning_dialog_guard_messages` in `tests/gui/test_app.py`, which exercises all three branches via a patched `QMessageBox` and asserts the correct message fires and the learning dialog is not opened.

```
$ QT_QPA_PLATFORM=offscreen uv run pytest tests/gui/test_app.py -q -p no:xvfb
4 passed
```

🤖 Generated with the help of Claude.

### DIFF
--- a/sleap/gui/app.py
+++ b/sleap/gui/app.py
@@ -1719,10 +1719,7 @@ class MainWindow(QMainWindow):
 
         if not self.state["filename"]:
             QMessageBox(
-                text=(
-                    "Please save your project before running training or "
-                    "inference."
-                )
+                text=("Please save your project before running training or inference.")
             ).exec_()
             return
 
--- a/sleap/gui/app.py
+++ b/sleap/gui/app.py
@@ -1708,7 +1708,25 @@ class MainWindow(QMainWindow):
             ).exec_()
             return
 
-        if not self.state["filename"] or self.state["has_changes"]:
+        if self.labels is None or len(self.labels.videos) == 0:
+            QMessageBox(
+                text=(
+                    "This project has no videos. Please add a video before "
+                    "running training or inference."
+                )
+            ).exec_()
+            return
+
+        if not self.state["filename"]:
+            QMessageBox(
+                text=(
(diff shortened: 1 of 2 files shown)


gitttt-1234 (approved):
Looks good to me!

PR #2766, opened 2026-06-09
feat(qc): GUI config setter for Label QC detectors (#2756)

tom21100227:
**Stacked on #2765** (base = `feat/qc-label-error-detectors-2756`). Part of #2756.

Makes the new detectors configurable from the Label QC panel, so the default-ON detectors can be dialed back / tuned per project and the experimental ones enabled.

## What
A new **"Detector Settings"** group in the QC widget (between the Sensitivity slider and the viz tabs):
- **Toggles** — flip / chimera / duplicate (default-ON); chain-order / missing-node (default-OFF).
- **Threshold spinboxes** — flip fraction (0.50), duplicate score (0.50), chain turn-angle (60°) + inversion rate (0.30), missing-node probability (0.90). Each detector's thresholds disable when its toggle is off.
- **Ordered chains** — a text box to declare ordered node-name chains for the chain-order detector (e.g. `TTI, Tail_0, Tail_1, Tail_2, TailTip`), one chain per line.

`QCAnalysisWorker` now takes a `QCConfig`; `_on_run_analysis` builds it from the controls (`_build_qc_config()`) and passes it to `LabelQCDetector`. The display-time **Sensitivity** slider is unchanged.

## Tests
**61 GUI QC tests pass** (11 new + 50 existing, 0 regressions); ruff clean. The panel was rendered offscreen and visually verified (reliable detectors checked with enabled thresholds; experimental rows unchecked with disabled thresholds).

## Follow-up
Ground-truth **calibration** of the default thresholds (the detectors over-flag on real heterogeneous data) is still pending — these controls are what make that tuning possible in-app.

Issue #2756, opened 2026-06-09
Label QC(enhancements): improve detection of common labeling errors (duplicates, L/R flips, chimeras, chain-order, occluders, missing points)


tom21100227:
Sub-issue of #2751.

> ⚠️ **Blocked by #2753 — do not implement until invisible-node handling is fixed.**
> These detectors build directly on the per-instance geometry + visibility signals. While invisible nodes are mis-handled (#2753), any new detector would be calibrated on incorrect visibility, so this whole effort should wait until #2753 is resolved.

This is the consolidated tracking issue for **improving what Label QC can actually detect** — i.e. the common labeling errors that the current engine misses. Tracked as one effort (not split into separate sub-issues per detector).

## How Label QC works today (context)

Label QC runs entirely on the **keypoint geometry of user-labeled instances** — it never reads image pixels, point confidence, or track identity (verified: no `scores=`, no pixel/track reads anywhere in `sleap/qc/`). Each instance becomes an **18-dim feature vector** (12 baseline + 6 "v3": edge/angle/pairwise z-scores, bbox/hull area & compactness, centroid spread, symmetry consistency, visibility rate/pattern, curvature, nearest-neighbor pose distance) → an unsupervised **GMM** (z-score fallback under 50 instances) → a 0–1 anomaly score. The GUI threshold slider only re-filters cached scores. Separately, **3 frame-level rules** run per frame: instance-count, negative-frame-with-instances, and duplicate detection (bbox-IoU + node-overlap). Backend `sleap/qc/` → `QCWidget`/`QCDockWidget` GUI → navigate to the frame.

**Three structural blind spots** this effort targets:
1. **Relational errors** (duplicate, chimera) — the GMM scores each instance *in isolation*, so it cannot represent relationships between instances or between node clusters.
2. **Chirality / ordering errors** (L/R flip, chain order) — every feature is built from distances + *unsigned* angles, which are invariant under reflection and nearly invariant under permutation.
3. **Appearance errors** (points on occluders, missing-but-visible) — no image pixels or model predictions are ever read.

> **Design principle:** feeding a single new signal into the GMM tends to *dilute* it (1 of 18+ dims, and the model self-trains on data that may contain the very errors we hunt). So most detectors below are best implemented as a **dedicated feature + a hard-rule override / deterministic check**, mirroring how duplicate detection already works alongside the GMM.

---

### (a) Duplicate / split instances on one animal — **Effort: moderate** — geometry-only (track-id optional)
[...]

PR #2768, opened 2026-06-09
feat(qc): appearance + in-sample prediction detectors (B2) (#2756)

tom21100227:
**Stacked on #2766** (base = `feat/qc-detector-gui-2756`). Completes the #2756 detector set (all six).

The two input-channel-heavy detectors, both opt-in **non-GMM channels** (default-OFF, like missing-node):

## Detectors
- **(e) Appearance / wrong-object** (`sleap/qc/features/appearance.py`) — a per-node image-patch appearance model; flags keypoints on visually anomalous pixels (e.g. on white bedding over dark fur). Decodes labeled frames via the sleap-io backend → `channel_scores["appearance"]` → "Appearance outlier".
- **(f Tier-2) In-sample model prediction** (`sleap/qc/insample_prediction.py`) — runs a trained sleap-nn model on the labeled frames and flags unlabeled nodes the model confidently localizes (the "labelable but skipped" signal, distinguishing skipped from truly occluded). One batched inference after the per-instance loop → `channel_scores["prediction"]` → "Model expects a labeled part here". **Import-safe** (lazy torch/sleap_nn import) and **self-skips** when no model path is set or the model/skeleton is unavailable.

## Wiring
- `QCConfig`: `use_appearance`/`appearance_patch_size`/`appearance_min_samples`; `use_insample_prediction`/`insample_model_path`/`insample_peak_threshold`/`insample_min_confidence`/`insample_device` — all default-OFF/safe.
- `detector.py`: appearance fitted at `fit()` (decodes frames) + scored in the score loop (frame decoded **once per labeled frame**); in-sample prediction as a single post-loop batched stage. `results.py`: two `CHANNEL_ISSUE_LABELS` entries.
- GUI: two more "Detector Settings" toggles + a **model-folder picker** for f-T2.

## Real-data validation (250-frame CVAT subset)
- **Appearance:** scored all 674 instances; **4 surfaced** as "Appearance outlier" (synthetic white-cotton-on-dark-fur → 0.98).
- **f-T2:** **real Leopard-model inference** (~2 s / 250 frames on CUDA) found **84** candidate skipped-but-labelable nodes.

⚠️ **Calibration caveat:** f-T2 confidences (~0.5–0.65) sit below the default 0.7 threshold, and the over-flagging GMM dominates `top_issue` via the `max(GMM, channel)` merge — so these channel flags need threshold/GMM **calibration** to surface by default. The detectors work; surfacing is the calibration follow-up (same theme as the rest of the detector set).

## Tests
**405 QC + GUI tests pass** (incl. real-frame appearance + mocked-inference f-T2 integration tests, and GUI control tests); ruff `check` + `format` clean. Default config (both OFF) leaves behavior unchanged and runs no inference.

### DIFF
--- a/sleap/gui/widgets/qc.py
+++ b/sleap/gui/widgets/qc.py
@@ -965,6 +965,50 @@ class QCWidget(QtWidgets.QWidget):
         )
         self._add_detector_row(5, self._cb_missing, missing_thr_row)
 
+        # --- Appearance / wrong-object (B2 channel), experimental, default-OFF ---
+        self._cb_appearance = QtWidgets.QCheckBox("Appearance / wrong-object")
+        self._cb_appearance.setChecked(False)
+        self._cb_appearance.setToolTip(
+            "Flag a keypoint placed on visually-wrong pixels (e.g. on bedding "
+            "instead of fur), using a per-node image-appearance model. "
+            "Experimental; off by default."
+        )
+        # No tunable hard threshold in the GUI: appearance is a channel score.
+        appearance_note = QtWidgets.QLabel("(image-based)")
+        appearance_note.setEnabled(False)
+        appearance_note.setToolTip(
+            "Appearance scoring reads image patches around each node; it has no "
+            "hard threshold to tune here."
+        )
+        self._add_detector_row(6, self._cb_appearance, appearance_note)
+
+        # --- In-sample model prediction (B2 channel), experimental, default-OFF ---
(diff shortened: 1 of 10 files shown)


tom21100227:
Superseded by #2770 — the full Label QC effort (#2756) is consolidated onto one branch (`feat/qc-2756-all`) for review + real-life testing, with the #2769 GUI revamp added on top.

Issue #2762, opened 2026-06-09
Master Issue: Label GUI Improvements


tom21100227:
Running list of all Label GUI improvements

PR #2771, opened 2026-06-09
fix(gui): spread occluded nodes via force-directed layout (kept hidden) when converting predictions (#2764)

### LINKED ISSUE #2764: Label GUI: Better keypoints initialization when converting predictions to instances
We have a great selection of instance initialization pattern : 

Image

but when we convert predictions back to labelled instances the keypoints are still intitialized randomly. badbad

tom21100227:
## Summary

When converting predicted instances to user instances (**Labels ▸ Add Instances from All Predictions on Current Frame** and **Accept All Predictions…**), nodes the model didn't detect come back as `xy=NaN` (e.g. the whole occluded side of the animal). They're correctly kept **hidden** (`visible=False`) — but a `NaN` coordinate renders at a garbage/default location when the user enables **"show non-visible nodes"** (`QtInstance` still creates a node item for every node in that mode, positioned at its `xy`). That's the "initialized in the wrong place" behavior.

This PR gives those hidden nodes a sensible **spread-out** position while keeping them hidden, so "show non-visible nodes" reveals them spread on the body — ready to drag out.

## What changed

`AddUserInstancesFromPredictions.fill_missing_predicted_nodes(new_instance)` positions each undetected node with a **force-directed (spring) layout** of the skeleton graph, centered on the detected keypoints' centroid and scaled to their extent, and leaves it `visible=False` / `complete=False`. Detected points, the track, and the `from_predicted` link are untouched.

- Spreads the occluded nodes by graph connectivity — always on the body, always distinct, **regardless of how many nodes were detected**.
- Needs no template/labels/player, so the single-frame, all-frames, and headless paths behave identically.
- No-op when every node was detected, or when nothing was detected (no anchor for the layout center).

## Why force-directed (and not template alignment)

An earlier revision aligned the average labeled pose ("best"/template, like Add Instance) onto the detected keypoints. On real refinement data that put **23% of instances' occluded nodes off the animal**: aligning to only a few detected keypoints is underdetermined (you can't infer where the tail is from just the nose), so it flips/flings. Measured on the same data, force-directed is **0% off-body** with every multi-missing instance fully spread.

## Tests

`tests/test_predictions_to_instances.py` — **30 passing**: missing nodes land spread (distinct) + hidden + near the detected node; per-node hidden/link-preserved; all-detected / nothing-detected / single-node no-ops; and `do_action` wiring (single-frame + all-frames). `ruff` clean.

## Manual check

Open a project with predictions that have occluded nodes → **Add Instances from All Predictions** → the converted instance looks like the prediction (occluded nodes hidden). Toggle **show non-visible nodes** → the hidden nodes now sit spread out on the body (not at the corner/origin, not clumped).

Part of #2762. Closes #2764.

🤖 Implemented with Claude Code (workflow + review/maintainer-feedback fixes).

### DIFF
--- a/sleap/gui/commands.py
+++ b/sleap/gui/commands.py
@@ -4817,6 +4817,66 @@ class AddUserInstancesFromPredictions(EditCommand):
 
         return new_instance
 
+    @staticmethod
+    def fill_missing_predicted_nodes(
+        context: CommandContext,
+        new_instance: Instance,
+        copy_instance: PredictedInstance,
+    ):
+        """Initialize undetected nodes on a converted prediction instance.
+
+        ``make_instance_from_predicted_instance`` copies the model's detected
+        keypoints and leaves any node the model did not detect at ``xy=NaN``
+        / ``visible=False``. This fills those missing nodes with sensible,
+        visible positions using the same machinery that initializes a
+        brand-new instance (``AddMissingInstanceNodes.add_best_nodes``):
+        the template is aligned onto the already-detected points (or, when
+        nothing was detected, centered on the current view), and any node
+        still missing falls back to random in-view placement. Filled nodes
+        are marked ``visible=True`` and ``complete=False`` so the user can
+        see and drag them; already-detected points, the track, and the
+        ``from_predicted`` link are left untouched.
+
+        This is a no-op when there is no GUI ``player`` available (e.g. the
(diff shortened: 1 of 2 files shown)


tom21100227:
looking better, but need to spread out some of those keypoints. one more commit and then it should be good to go.

Issue #2808, opened 2026-07-01
GUI "Entire video" inference emits fragile `--frames 0,-N` and can request an out-of-range frame


tom21100227:
## Summary

Running inference from the GUI (**Predict → Run Inference**) with the frame range set to **"Entire video"** builds a command containing `--frames 0,-N` (e.g. `--frames 0,-10245`). This encoding is fragile and can request a frame index past the end of the video, crashing inference with `IndexError: Failed to read frame index N`.

Reported in discussion #2807.

## Where it comes from

- `MainWindow._get_frames_for_prediction` (`sleap/gui/app.py`) encodes "Entire video" as the half-open range `[0, len(video))`, stored as `[0, -len(video)]` via the local `encode_range` helper.
- `VideoItemForInference.cli_args` (`sleap/gui/learning/runners.py`) serializes that to `--frames 0,-N`.
- `sleap track` (→ sleap-nn) only decodes `0,-N` back into a range because `sleap_nn`'s `frame_list` does `min_max[0].rstrip(",")`. That is an undocumented legacy quirk; if it were ever simplified, the GUI would silently pass the literal two-element list `[0, -N]`.

## The problem

Pinning the "Entire video" endpoint to the GUI's frame count has two downsides:

1. **Fragile / confusing encoding.** `--frames 0,-N` reads like the two-frame list `[0, -N]` (which is how the discussion author interpreted it) and only works as a range thanks to the `rstrip` quirk above.
2. **Off-by-one / out-of-range.** The endpoint is the GUI's `len(video)`. If that count is one greater than the number of frames the inference video reader can actually decode — e.g. the 1-indexed vs 0-indexed discrepancy in #2790, or a video whose container metadata over-reports its frame count — inference iterates to a non-existent frame and raises `IndexError: Failed to read frame index N`.

## Reproduction (from discussion #2807)

1. Open a project with a video.
2. Predict → Run Inference; configure models + tracking.
3. Set the frame range to "Entire video"; Run.

Observed generated command and error (paths redacted):

```
sleap track --gui --data_path .../project.slp --video_index 0 --frames 0,-10245 --model_paths ... -o .../output.predictions.slp ...
```
```
IndexError: Failed to read frame index 10245.
```

## Fix

For the entire-video selection, omit `--frames` and let the inference CLI predict on all frames using its own frame count (its documented default when `--frames` is not given). This removes the fragile `0,-N` encoding and decouples the run from the GUI's frame count.

[...]

PR #2848, opened 2026-08-12
fix(gui): run "Predict on entire video" on the video file, not the project .slp

gitttt-1234:
## Summary

- "Run Inference → entire video" silently capped at however many frames were **already labeled** in the project, because the GUI pointed `--data_path` at the project `.slp` instead of the video file.
- `VideoItemForInference` now targets the video's own file (and drops `--video_index`) whenever the video *has* a file of its own, so `sleap-nn` reads every requested frame via `VideoProvider` instead of subsetting labeled frames via `LabelsProvider`.
- Embedded (`.pkg.slp`) and image-sequence videos have no separate video file, so they keep the project + `--video_index` route — which is also the only correct source for them.

## The bug

`VideoItemForInference` is the item used for "inference on an arbitrary list of frames from a video" (entire video, clip, current frame, random sample). Its `path` returned `labels_path` whenever one was set — and `ItemsForInference.from_video_frames_dict` always sets it — so the CLI became `--data_path <project.slp> --video_index N --frames 0,-1499`.

On the `sleap-nn` side, `.slp` + `--video_index` + `--frames` routes through `LabelsProvider`, which can only *subset frames that already have a `LabeledFrame`*; it never synthesizes frames for indices that were never labeled. So "predict on frames 0–1499" against a `.slp` with 20 labeled frames can never yield more than those 20, no matter what range is requested.

Repro (`sleap-nn 0.3.3`), on a project with 20 labeled frames of a 1500-frame video:

```
sleap predict --gui --data_path .../test_inference.slp --video_index 0 --frames 0,-1499 \
  --model_paths .../260428_082921.centroid.n=2560 \
  --model_paths .../260428_113349.centered_instance.n=2560 -o .../predictions.slp

Starting inference | source=LabelsProvider | frames=20 | ...
```

1500 frames requested, 20 processed. Detection quality was never the problem (2.00 instances/frame) — the run was scoped to the wrong frame set.

`DatasetItemForInference` (the user-labeled / suggested / predicted filters) is genuinely annotation-status based and does **not** have this bug; it is untouched here.

## Changes Made

`sleap/gui/learning/runners.py`:

- **`VideoItemForInference.uses_labels_path`** (new): true only when the video has no single video file to point `--data_path` at — an embedded `.pkg.slp` (where `video.filename` *is* a `.slp`) or an image-sequence video (where `filename` is a `list`). Those cases need `--video_index` to scope the run.
- **`VideoItemForInference.path`**: returns the video's filename unless `uses_labels_path`. Remote URLs are passed through verbatim (`os.path.abspath()` would rewrite `https://h/v.mp4` into `<cwd>/https:/h/v.mp4`); relative paths need no handling because `sleap_io.load_slp` already resolves them against the project location and rewrites `filename`.
- **`VideoItemForInference.cli_args`**: emits `--video_index` exactly when `--data_path` is the project `.slp`. Also omits `--frames` when the frame list is empty instead of raising `TypeError` — "no frames" means all frames, which is what omitting the flag does, and matches what the `frames` attribute's docstring already promised.
- **`InferenceTask.make_predict_cli_call`**: output filenames are derived from `os.path.basename(item.path)`, which now carries the video name directly. The old `video_name_prefix` (from `--video_index` → `labels.videos[i].filename`) would have duplicated it (`video_1.mp4_video_1.mp4…`). The prefix is kept only for the `.slp`-scoped case, keyed on the **index** (`video3_train.pkg.slp.<ts>.predictions.slp`) — every video in a `.pkg.slp` reports the same filename, so the old name-based prefix disambiguated nothing there.

## API Changes

- `VideoItemForInference.path` no longer returns `labels_path` for videos that have their own file. `labels_path` and `video_idx` are still populated and are still used for embedded/image-sequence videos and for output naming.
- `VideoItemForInference.cli_args` no longer emits `--video_index` for file-backed videos.
- New public property `VideoItemForInference.uses_labels_path`.
- Auto-generated prediction filenames change: `video_1.mp4_project.slp.<ts>.predictions.slp` → `video_1.mp4.<ts>.predictions.slp` (and `video3_train.pkg.slp.<ts>.predictions.slp` for embedded videos). Explicit `-o` paths are unaffected.

## Testing

11 new tests in `tests/gui/learning/test_cli_construction.py` (48 total in the file), plus 3 existing assertions inverted to the corrected behavior. Full `tests/gui`: 881 passed, 2 skipped.

Verified end-to-end against real projects and models, not just unit-level CLI construction:

| Video backend | `--data_path` | `--video_index` | Live check |
|---|---|---|---|
| `MediaVideo` (`.mp4`) | the video file | no | 20 labeled / 1500 frames → `source=.../video_1.mp4 \| frames=30` for `--frames 0,-29` (was `frames=0`) |
| `HDF5Video`, plain (`.h5` + dataset) | the `.h5` | no | 3 labeled / 40 frames → `source=VideoProvider \| frames=40` |
| `HDF5Video`, embedded (`.pkg.slp`) | project `.slp` | yes | 30-video training package, per-video scoping intact |
| `ImageVideo` (1 or N images) | project `.slp` | yes | real `sio.Video` objects |
| Remote URL | the URL verbatim | no | unit test |

Also confirmed: `InferenceTask.merge_results` attaches the video-sourced predictions to the existing project video (video count unchanged, no duplicate video added); `--only_labeled_frames` runs still log `source=LabelsProvider | frames=20` exactly as before; `labels.videos.index(video)` is identity-based, so index scoping is correct even with 30 identically-named embedded videos; channel handling comes from the model's preprocess config (`ensure_rgb`/`ensure_grayscale`), so switching providers cannot change it.

## Design Decisions

- **Gate on "does this video have its own file", not on "was a `labels_path` supplied".** An embedded video's `filename` is the `.pkg.slp` itself, so switching it to the "video path" would hand `sleap-nn` a `.slp` *without* `--video_index` — running unscoped across every video in the project. For embedded videos the frames that exist are exactly the ones stored in the project, so `LabelsProvider` is the right source.
- **Index-based output prefix for the `.slp` case.** After this change that branch is only embedded/image-sequence videos, which all share one filename; the video index is the only discriminator that actually disambiguates per-video runs.
- **`DatasetItemForInference` left alone.** It selects frames by annotation status, which is exactly what a `.slp` `--data_path` expresses.

## Future Considerations / Known Limitations

Out of scope here, each pre-existing or inherent:

1. **Image-sequence videos still cap at labeled frames.** `sleap-nn`'s `--data_path` has no directory or image-list form, so there is no video route to give them.
2. **Entire-video mode on an embedded `.pkg.slp` yields 0 frames.** The GUI encodes the range as `[0, len(video))` over *embedded positions*, while the labeled frames carry their original source indices (e.g. `6760, 7436, …`), so the two index spaces don't intersect. Verified: `--frames 0,-79` on a pkg → `frames=0`. Needs an embedded-position ↔ source-`frame_idx` mapping (or restricting embedded videos to annotation-status targets).
3. **"Skip user labeled" (`--exclude_user_labeled`) is inert on the raw-video route** — `sleap-nn` warns and runs all frames. Entire-video mode encodes frames as a `[0, -N]` range, so it can't be filtered GUI-side without expanding to a huge explicit index list; a proper fix needs annotation-status filtering for a video source in `sleap-nn`.
4. **Two videos with the same basename in different directories** can collide on one output filename within the same second. Equally true before this change (the old prefix was also just the basename).
5. A companion `sleap-nn` defensive-UX fix (warn on out-of-range `--frames` for a `.slp` source) is tracked separately.

## Related Issues

- #2808 — same code path (`VideoItemForInference.cli_args`), fragile `--frames 0,-N` encoding that can request an out-of-range index. Not fixed here, but this PR is a prerequisite for its proposed fix: `cli_args` now omits `--frames` when the frame list is empty, so "entire video → no `--frames`, let the reader decide" becomes a selection-encoding change only. Stays open.
- Discussion #2844 ("Find instances on suggestion but not on entire video") — same symptom shape. Its root cause was assessed separately as the `sleap-nn` centroid/centered-instance scale-sharing bug (fixed in `sleap-nn` 0.3.3); the scoping defect fixed here may also have contributed. Not claimed as closed.

Full investigation notes live in the `sleap-nn` repo under `scratch/2026-08-12-gui-entire-video-frame-cap/`.

---
🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/sleap/gui/learning/runners.py
+++ b/sleap/gui/learning/runners.py
@@ -430,14 +430,15 @@ class VideoItemForInference(ItemForInference):
     This allows for inference on an arbitrary list of frames from video.
 
     Attributes:
-        video: the :py:class:`Video` object (which already stores its own path)
-        frames: list of frames for inference; if None, then all frames are used
-        use_absolute_path: whether to use absolute path for inference cli call
         video: The :py:class:`Video` object (which already stores its own path)
         frames: List of frames for inference; if None, then all frames are used
-        labels_path: Path to .slp project; if None, then use video path instead.
-        video_idx: Video index for inference; if None, then first video is used. Only
-            used if labels_path is specified.
+        use_absolute_path: Whether to use absolute path for inference cli call
+        labels_path: Path to the .slp project the video belongs to, if any. Used as
+            `--data_path` only when the video has no video file of its own to point at
+            (see `uses_labels_path`); a video with its own file is inferred on
+            directly, so unlabeled frames aren't skipped.
+        video_idx: Index of this video within the project's videos. Passed as
+            `--video_index` only when `--data_path` is the project .slp.
     """
 
     video: Video
@@ -447,18 +448,55 @@ class VideoItemForInference(ItemForInference):
     video_idx: int = 0
 
     @property
+    def uses_labels_path(self) -> bool:
+        """Whether inference has to run on the project rather than a video file.
+
+        True only when the video has no single video file of its own to point
+        `--data_path` at:
+
+        * Embedded projects (`.pkg.slp`): the images live inside the project file, so
+          `video.filename` *is* a `.slp` (with `backend.dataset` naming the embedded
+          video). This is also the only correct source -- the frames that exist for an
+          embedded video are exactly the ones stored in the project.
+        * Image-sequence videos: `filename` is a list of image paths, which is not
+          something a single `--data_path` can express.
+
+        Both cases need `--video_index` to scope the run to this video. Falls back to
+        the video file when there is no project path to use.
+        """
+        if not self.labels_path:
+            return False
+        filename = self.video.filename
+        if not isinstance(filename, (str, Path)):
+            # Image sequence: a list of per-frame image paths.
+            return True
+        return str(filename).endswith(".slp")
+
+    @property
     def path(self):
-        if self.labels_path is not None:
+        # Prefer the video's own file: this item runs on an arbitrary list of frame
+        # *indices*, most of which are typically unlabeled (e.g. "entire video"). A
+        # .slp `--data_path` routes sleap-nn through `LabelsProvider`, which can only
+        # subset the frames already labeled in the project, so the run would silently
+        # cap at the labeled-frame count. Pointing at the video instead uses
+        # `VideoProvider`, which reads any frame index.
+        if self.uses_labels_path:
             return self.labels_path
-        if self.use_absolute_path:
-            return os.path.abspath(self.video.filename)
-        return self.video.filename
+        filename = self.video.filename
+        # Remote videos are passed through verbatim; `abspath` would rewrite a URL
+        # into a cwd-relative path. Local relative paths are already resolved
+        # against the project location by `sleap_io.load_slp`.
+        if self.use_absolute_path and "://" not in str(filename):
+            return os.path.abspath(filename)
+        return filename
 
     @property
     def cli_args(self):
         arg_list = list()
         arg_list.extend(["--data_path", f"{self.path}"])
-        if self.labels_path is not None:
+        # `--video_index` picks one video out of a `.slp`; it has no meaning when
+        # `--data_path` is the video file itself (there is only one video then).
+        if self.uses_labels_path:
             arg_list.extend(["--video_index", str(self.video_idx)])
 
         # TODO: better support for video params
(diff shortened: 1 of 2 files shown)


WeissShahaf:
installing from source fixed it , but now excluding user labels doesn't work:

Running: sleap predict --gui --data_path //gpfs.corp.brain.mpg.de/stem/data/Project_Stop/Behaviour experiments/00_data/260807_AFM-18752/AFM-18752_camera_bottom2.avi --frames 0,-14999 --model_paths \\gpfs.corp.brain.mpg.de\stem\data\Project_Stop\sleap\bottom\models\260819_183139.single_instance.n=444 -o //gpfs.corp.brain.mpg.de/stem/data/Project_Stop/sleap/bottom\predictions\AFM-18752_camera_bottom2.avi.260821_111304.predictions.slp --batch_size 32 --exclude_user_labeled
W0821 11:13:30.504000 25008 Lib\site-packages\torch\utils\flop_counter.py:29] triton not found; flop counting will not work for triton kernels
2026-08-21 11:13:35 | --exclude_user_labeled require a `.slp` --data_path (they filter frames by annotation status); the given source has no annotations to filter on, so these flags have no effect here. Running on all frames. Use --frames to subset a video by index instead.
2026-08-21 11:13:35 | sleap-nn 0.3.3 | Python 3.12.12 | PyTorch 2.11.0+cu128 | CUDA 12.8 | 1 GPU(s)
2026-08-21 11:13:37 | Loaded inference model | type=single_instance | backbone=unet | nodes=7 | device=cuda | batch_size=32 | peak_threshold=0.2 | max_instances=None | integral_refinement=integral | paf_workers=0
2026-08-21 11:13:37 | Starting inference | source=\\gpfs.corp.brain.mpg.de\stem\data\Project_Stop\Behaviour experiments\00_data\260807_AFM-18752\AFM-18752_camera_bottom2.avi | frames=15000 | videos=1 | shape=876x2048x1 | fps=50.0 | tracking=False

